Issue #3, opened 2016-03-01
Fix error handling in Render


synlestidae:
This code sample, when placed in tests/templates/ (and added to the test suite)

`<h1>{{object.nonexistent_field}}</h1>`

causes the following error

`thread 'test_templates' panicked at 'called 'Option::unwrap()' on a 'None' value', ../src/libcore\option.rs:330`

It's pretty clear why -- render.rs is full of unsafe unwraps. @Keats you already know about this, but I thought I should file the first issue to track this.

synlestidae:
To start, lookup_variable now returns an Option type: 886850369249741facdd820c5035b94c18b391fb. I expect that soon render will be able to capture when a variable lookup failed, and return an error like "so-and-so variable not initialized".

Keats:
Thanks for that!
Lots of work to be done with error handling!

Keats:
Error handling for rendering is now in master

PR #5, opened 2016-03-06
Handle unresolvable keys

stephanbuys:
This is, perhaps, the start of handling missing keys without resorting to panic'ing the thread.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,22 +1,26 @@
 [package]
+authors = ["Vincent Prouillet <vincent@wearewizards.io>"]
 name = "tera"
 version = "0.1.0"
-authors = ["Vincent Prouillet <vincent@wearewizards.io>"]
 
 [dependencies]
-walkdir = "0.1"
+env_logger = "0.3.2"
 serde = "0.7"
 serde_json = "0.7"
 serde_macros = "0.7"
-clippy = {version = "~0.0.35", optional = true}
+walkdir = "0.1"
 
+[dependencies.clippy]
+optional = true
+version = "~0.0.35"
 
 [features]
 default = []
 dev = ["clippy"]
 
+[profile]
 
 [profile.bench]
-lto = true
 codegen-units = 1
+lto = true
 opt-level = 3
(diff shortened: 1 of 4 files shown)


stephanbuys:
Oh, and also a 'once' initialiser that allows specifying a single file, as opposed to walking a tree by default.

Keats:
Hey thanks for the PR.

For error handling, it seems @synlestidae started with #3 using Option which is imo the right way to do it: `render` should return a Result and it shouldn't be possible to ignore it

For the once initialiser, could you expand on why you would want a single file rather than a directory? Loading only one file will mean that inheritance won't work with it so it could lead to users confusion since they might expect Tera to find the parent file, which it won't do (and probably will never do) and i'm a bit reluctant to put API that can't use 100% of tera capabilities

stephanbuys:
@Keats, full disclosure, I'm a complete novice :smile: (but I'm willing to help if I can) , so if nothing else let me lay out some observations.

I agree it should return a Result, with the Err containing a hint why the render failed, but I would also argue for a 'ignore' mode, which the Jinja2 library supports, it kind of feeds into the second part...

There are other non-html consumers of Jinja2 such as Saltstack (and some of our own tools), which applies the power of it's template engine to non-html files. In our case we have template files marked with a special extension, in my tests I was '`glob`bing' for our template files and then just '`once`'ing the individual files, but maybe a better solution is to use "Glob" instead of "WalkDir", that way the user can achieve all of the power by using for example, `./templates/**`, or if the want to be specific `./templates/**/*.tpl` - I've used crate `glob` and it works nicely https://doc.rust-lang.org/glob/glob/index.html

stephanbuys:
Just another note, I think @synlestidae 's is definitely on the right track, so please feel to close this pull request, perhaps just consider the `glob` comment.

Keats:
I like the globbing aspect, created #8 to remember it. Would globbing avoid the need for the `once` method?

stephanbuys:
@Keats yeah, with globbing you would be able to specify just one file if need be, I think it would cover most use-cases I can think of now. (Apologies, long winded way to say, yes - wont need 'once' anymore)

PR #6, opened 2016-03-07
Change render function on Render to return Result

synlestidae:
Fixed a portion of the unwraps in render.rs for #3 

Have more changes to make before that is solved :)

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -355,25 +355,25 @@ mod tests {
 
     #[test]
     fn test_render_simple_string() {
-        let result = Template::new("", "<h1>Hello world</h1>").render(Context::new());
+        let result = Template::new("", "<h1>Hello world</h1>").unwrap().render(Context::new());
         assert_eq!(result, Ok("<h1>Hello world</h1>".to_owned()));
     }
 
     #[test]
     fn test_render_missing_variable_error() {
-        let result = Template::new("", "<h1>{{ product }}</h1>").render(Context::new());
+        let result = Template::new("", "<h1>{{ product }}</h1>").unwrap().render(Context::new());
         assert!(result.is_err());
     }
 
     #[test]
     fn test_render_missing_variable_and_field_error() {
-        let result = Template::new("", "<h1>{{ product.nonexistent_field }}</h1>").render(Context::new());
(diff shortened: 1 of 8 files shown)


src/render.rs
```diff
@@ -35,28 +35,34 @@ impl ForLoop {
...
-    output: String,
     context: Json,
     ast: Node,
-    for_loops: Vec<ForLoop>
+    for_loops: Vec<ForLoop>,
+    render_error: Option<RenderError>
 }
 
+#[derive(Debug, Eq, PartialEq)]
+pub struct RenderError {
```
Keats: can you put the errors in a separate file and maybe use https://github.com/tailhook/quick-error or structure them like https://github.com/kbknapp/clap-rs/blob/master/src/errors.rs ?

src/render.rs
```diff
@@ -275,50 +296,57 @@ impl Renderer {
...
             i += 1;
         }
+        Ok(output)
     }
 
-    pub fn render_node(&mut self, node: Node) {
+    pub fn render_node(&mut self, node: Node) -> Result<String, RenderError> {
         match node.specific {
-            Text(ref s) => self.output.push_str(s),
+            Text(ref s) => Ok(s.to_string()),
```
Keats: `to_owned` is preferrable to `to_string`

tests/templates/basic_space.html
```diff
@@ -0,0 +1 @@
+<h1>{{object.doesnt_exist}}</h1>
```
Keats: name of the file seems wrong?

Keats:
Looks good for a start!
Note i'll be away for 2 weeks without laptop on wednesday so i won't be commenting too much (or at all) / merging during that time

synlestidae:
Okay that's cool! 

I hope that's all the feedback you have for now. I'll implement it later today when I have the project open

Keats:
I had to change the render a bit to add inheritance (see #12) but that shouldn't be a big change

Keats:
@synlestidae do you think you will have time to work on it this week? i'd like to release a 0.1 on crates.io asap.
Otherwise I can continue from your pr

Keats:
Closed in favour of #16

PR #9, opened 2016-03-07
Use Glob instead of WalkDir, Add logging

stephanbuys:
This changes WalkDir into using Glob. Also adds logging.

Unsolved is the removal of the directory for convenience.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,22 +1,26 @@
 [package]
+authors = ["Vincent Prouillet <vincent@wearewizards.io>"]
 name = "tera"
 version = "0.1.0"
-authors = ["Vincent Prouillet <vincent@wearewizards.io>"]
 
 [dependencies]
-walkdir = "0.1"
+glob = "0.2.11"
 serde = "0.7"
 serde_json = "0.7"
 serde_macros = "0.7"
-clippy = {version = "~0.0.35", optional = true}
+# walkdir = "0.1"
 
+[dependencies.clippy]
+optional = true
+version = "~0.0.35"
 
 [features]
 default = []
 dev = ["clippy"]
 
+[profile]
 
 [profile.bench]
-lto = true
 codegen-units = 1
+lto = true
 opt-level = 3
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,11 +4,15 @@ name = "tera"
 version = "0.1.0"
 
 [dependencies]
+log = "^0.3"
 glob = "0.2.11"
 serde = "0.7"
 serde_json = "0.7"
 serde_macros = "0.7"
 # walkdir = "0.1"
+
+[dev-dependencies]
+env_logger = "^0.3"
 
 [dependencies.clippy]
 optional = true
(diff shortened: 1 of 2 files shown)


Keats:
Ah sorry ended up doing it in https://github.com/Keats/tera/commit/7b15024910ffa9be3e2e1b31f562afacf3616177

I'll still do a code review

Cargo.toml
```diff
@@ -1,22 +1,30 @@
 [package]
+authors = ["Vincent Prouillet <vincent@wearewizards.io>"]
 name = "tera"
 version = "0.1.0"
-authors = ["Vincent Prouillet <vincent@wearewizards.io>"]
 
 [dependencies]
-walkdir = "0.1"
+log = "^0.3"
+glob = "0.2.11"
```
Keats: if you use "0.2" in cargo.toml, it will update if "0.2.12" is released automatically which is a bit nicer

src/lib.rs
```diff
@@ -51,18 +53,22 @@ impl Tera {
         let mut templates = BTreeMap::new();
 
         // We are parsing all the templates on instantiation
-        for entry in WalkDir::new(dir).into_iter().filter_map(|e| e.ok()) {
-            let path = entry.path();
+        for entry in glob(dir).unwrap().filter_map(|e| e.ok()) {
+            let path = entry.as_path();
             // We only care about actual files
             if path.is_file() {
+                debug!("Found file: {:?}", path);
```
Keats: you marked env_logger as dev dependencies but this would be run in release as well and doesn't the logger need to be initialised?

stephanbuys: I modelled it after how hyper uses the env_logger. So, the program using tera would do the initialisation if they are interested in using debugging. Otherwise it should have no effect, but I agree that the dependency should be not just be for dev, I'll refactor.

stephanbuys:
:smile: All good, just one comment, is the test for glob `lib.rs:53` necessary? Can't we trust the user to decide to forgo 'inclusion, etc'. I'm sure most end-users would end up using a `*` but some of my use-cases don't need it.

Keats:
The issue i have with logging is that I don't want my dependencies to pollute my log output/files, error messages should explain everything and be enough. If it's a bug in the library println is good enough for debug right now.

For the `*` test, keep in mind the Tera constructor should only be called once and most projects have more than 1 template (I mean 99.9% of them have more than 1)

stephanbuys:
Hi, agreed with all points above, I referred to the hyper library for `debug!` usage, agreed that it shouldn't pollute, but used strategically it could provide valuable insights when things go wrong.

As to the `*` request, I understand, but what is the harm :smile: of allowing the edge case, at the cost of abbreviation (line `63` onward), in my case I also use 99% glob patterns, but would like to allow 'one shot' template handling. Pure request at very little actual cost, but I won't pursue it any further if you decide otherwise. 

Example:
```
  let parent_dir = match dir.find("*") {
    Some(x) => {
      dir.split_at(x).0
    },
    None => ""
  };
```

Keats:
I think i'll keep it that way for now to ensure a very simple API with no gotcha for now but we can definitely revisit later, feel free to open an issue for that so we don't forget

PR #11, opened 2016-03-08
WIP: Add a token for LStrip scenarios

stephanbuys:
Potentially useful, but also just for comment/review (never worked with parsers before).


Keats:
I want that to be the default behaviour so no need to add new tokens (since there is no particular reasons to want to keep the tag whitespace afaik)

PR #13, opened 2016-03-26
Clean HTML output

Keats:
The only issue seems to be merging a child with
a parent template which is tricky.

### DIFF
--- a/src/nodes.rs
+++ b/src/nodes.rs
@@ -76,11 +76,20 @@ impl Node {
         }
     }
 
-    // Only used by SpecificNode::List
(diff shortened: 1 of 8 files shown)


Keats:
@stephanbuys is that enough in terms of whitespace control?
I'm not sure what other template engines do or what's needed, I never really used that feature in jinja/django

stephanbuys:
@Keats, we use the `{%- for ... -%}` and `{%- if ... -%}` quite extensively. I tested this PR and I don't think it addresses it.

PR #26, opened 2016-04-29
Comments block

Alex-PK:
Hi,

sorry for the "unrequested" PR. I discovered Tera a few days ago, and always wanted to learn more about lexers and parsers. I thought that helping implement comments in tera would have been an easy pick to start with, and, as I would like to switch to Tera from Handlebars in a project of mine, catch three birds with a stone. :)

I implemented them in two different ways. This is the "clean" one, transforming comments into delimited blocks for the parser, comprising of an opening tag, a single "comment body" tag and a closing tag.

I will send shortly a different patch, transforming the whole comment block into a single tag that in turns gets ignored by the parser.

If you find them useful, please go ahead and require changes or merge one of them. If you don't, discard them, no offense taken. :)

Have a nice day.


Keats:
Thanks I will have a look in a bit.

Since you're interested in using Tera, can you have a look at #24 and give your opinion?

src/parser.rs
```diff
@@ -202,6 +203,12 @@ impl Parser {
         self.add_node(node);
     }
 
+    fn parse_comment_block(&mut self)  {
+        let _ = self.expect(TokenType::CommentStart);
+        let _ = self.next_token();
```
Keats: can you run clippy? you'll need the nightly version of rust, do a `cargo update` and a `cargo build --features "dev"`.

You shouldn't need the `let _` bit

Keats:
lgtm, a few couple things:

- add an integration test (or add a comment to one of the existing one) in the tests directory
- make it so the comment existence is erased from the html output, ie if my comment is on line 1/char 1 and my html starts on line 2/char 1 then the output should have the html starts at line 1/char 1
- run clippy

src/parser.rs
Alex-PK: Ran it, but it didn't complain.
I removed the `let _`s anyway.

Alex-PK:
Integration test added.
As for the erasing, I would be unsure on how to do it. For example:

```html
<div>
  {# comment #}
  <h1>...</h1>
</div>
```

Should it be

```html
<div>
    <h1>...</h1>
</div>
```

or

```html
<div>
  <h1>...</h1>
</div>
```

Also: the detection of LF after a comment would need to be made in the lexer, as LF is normally grouped with spaces. If we detect a Space in the parser we would remove the LF and all indentation after it, not just the LF and, as I said above, we would keep the indentation of the comment, added to that of the following tag. But if we remove spaces before the comment we could "break" spacing when a comment is embedded in the middle of a line.

Keats:
It should be the second one.
I also just realised that we don't actually use the `position` on the Node struct at all and I can't remember why I put it. Hopefully it will come back to mind or i'll remove it.

Could you add the case you mentioned (and any other you can think of) on the integration test so we cover all the bases?

Keats:
@Alex-PK if you can weight on #23 i'd be interested of your input as well! (regarding my last comment)

Alex-PK:
I'm sorry, but I am unable to implement the white-space removal without changing and complicating both the lexer and the parser a lot, and still with the risk of removing the wrong ones in some cases. :(

Keats:
No worries, it's not a vital feature anyway, i'll have a look later. Can you just add all the cases you mentioned in the integration test? I'll merge it after that.
Also feedback on #33 is wanted :+1:

Keats:
@Alex-PK ping

Alex-PK:
Sorry, had a conference and a motherboard death just after that. :(
I changed it to ignore a \n after a comment, and added a (passing) integration test for my implementation.

src/lexer.rs
```diff
@@ -282,14 +286,37 @@ impl Lexer {
...
+                    BlockType::Comment => {
+                        StateFn(Some(lex_inside_comment))
+                    },
+                    _ => StateFn(Some(lex_inside_block))
+                }
+            },
+            DelimiterSide::Right => {
+                match self.current_block_type {
+                    BlockType::Comment => {
+                        if !self.is_over() && self.peek() == '\n' {
```
Keats: that condition seems a bit too specific, i think it's better to remove it

src/render.rs
```diff
@@ -430,6 +430,27 @@ mod tests {
...
+        assert_eq!(result.unwrap(), "<h1>Hello  world</h1>".to_owned());
+    }
+
+    #[test]
+    fn test_ignore_variable_in_comment() {
+        let mut context = Context::new();
+        context.add("name", &"Vincent");
+
+        let result = Template::new("", "My name {# was {{ name }} #} is No One.").render(context, HashMap::new());
+        assert_eq!(result.unwrap(), "My name  is No One.".to_owned());
```
Keats: The girl's name is Arya

Keats:
Thanks!

PR #27, opened 2016-04-29
Implemented comments by transforming them into an unused element.

Alex-PK:
Hi,

as I said in the previous PR, this is the second version of the comments lexer-parser.
This one finds the starting {# and lexes until the next #} comment in a single tag that then gets ignored by the parser.


Keats:
Closing this as I think the other PR is a cleaner solution

Issue #18, opened 2016-04-06
Filters


Keats:
Basic ones: capitalize, uppercase, lowercase, length, first, last, slugify, striptags, truncatechars, truncatewords, raw

Harder: 
- date/time formatting (use chrono?)
- pprint to pretty print a variable, useful for debugging. Having a magic argument that prints the full context would be very nice too

Keats:
Might be better as functions #23

PR #33, opened 2016-05-15
Functions

Keats:
Solves #23 and part of #18 

This is a test to see how Tera would look with functions replacing filters and templatetag (minus the ones taking a body that won't be supported for now).

Here are 2 examples of functions written for the tests:

```rust
fn format_price(args: Vec<Json>, kwargs: HashMap<String, Json>) -> TeraResult<String> {
    let price: i32 = from_value(args.get(0).cloned().unwrap()).unwrap();
    // Defaults to €
    let currency: String = kwargs.get("currency").map(|c| from_value(c.clone()).unwrap()).unwrap_or("€".to_owned());
    let do_conversion: bool = kwargs.get("do_conversion").map(|d| from_value(d.clone()).unwrap()).unwrap_or(false);

    if do_conversion {
        // imagine you're converting the price which defaults to € to £ for example
        // instead of just showing the price with another currency.
    }

    // In practice you would actually format the price
    Ok(format!("{} {}", currency, price))
}

// Simple fn
// TODO: error handling
fn uppercase(args: Vec<Json>, _: HashMap<String, Json>) -> TeraResult<String> {
    let text = args.get(0).cloned().unwrap();
    let str_text: String = from_value(text).unwrap();

    Ok(str_text.to_uppercase())
}
```
And you register them like so:

```rust
let mut tera = Tera::new("tests/templates/**/*");
tera.add_function("uppercase", uppercase);
tera.add_function("format_price", format_price);
```
It's missing error handling and a few helper methods but the important stuff is there.
The `uppercase` function in a template would be called `{{ uppercase(username) }}` rather than `{{ username | uppercase }}` like in Jinja2 or django. Piping results to other filters/functions is not implemented in this PR yet.

The `format_price` function can be called like `{{ format_price(product.price, currency=currency, do_conversion=false) }}` in a template. The function can also provide default values for the kwargs so `{{ format_price(product.price) }}` could still be a valid call.

The main downside of the implementation is that I had to start using serde::Value everywhere in the code which makes it harder to be typesafe (I guess). One solution would be to just dump strings for args/kwargs and let the function parse into the type it wants, which wouldn't work if you are passing a struct.

Another one is that I'm not too sure how to handle functions with a body, like:
```rust
{% call cache_block('currency', 'country') %}
...
{% endcall %}
```
as it would mean having to expose the renderer and passing the context as well so that would be a different thing than the functions in this PR

The code is still messy, mostly looking for feedback on how it looks to make sure it's the right direction


Keats:
cc @untitaker @Alex-PK @mitsuhiko @alexander255 @daogangtang

daogangtang:
looks some strange. let me see.

Keats:
Small ping

untitaker:
I'd have to try Tera out in my actual application, but that takes quite a while.

Alex-PK:
Looks good.

Maybe you are aware of the limitations, but this is what I notice with this implementation:

1- You can't have nested functions: substr(uppercase(mystring), 10) would not pass the lexer as it would close the substr after the first ')'.

2- Maybe I misunderstood the code, but apparently positional parameters and named ones are completely separated, and it's the function's responsibility to fallback to positional if named is not present. Given you already departed from jinja's syntax, I wonder if it makes sense to only have named parameters.

As for block functions, the way twig "solves" it is through macros. You first define a macro:

{% macro f(par1, par2) %}
 <h1>{{ par1 }} <small>{{ par2 }}</small></h1>
{% endmacro %}

and then call it:

{{ f('My title', 'with a note') }}

In the end, I think it's a different problem to be tackled in a different way.

Keats:
Thanks!

1. That will work in the final version, didn't bother making it work for the poc but i'll add it if people like the idea of functions
2. I thought about having only named parameters actually, but it might make simple calls a bit too verbose like `{{ capitalize(text=blabla) }}` vs `{{ blabla | capitalize }}` in jinja2 so not 100% convinced yet

Block functions are different from macros, which is another part that needs to be added. In my `cache_block` example above, it would look up in redis or something similar if we have a bunch of html for that tuple of currency/country and return that if yes, otherwise parse the block normally. That's kind of a niche thing though and it would probably be better to do that in Rust instead anyway.

daogangtang:
block function is really different with macro here. we need both of them.  
nested functions, i think is enough to transform value.
and i need macro to implement third web components to plug in.

daogangtang:
and if implement macro, we need design a method to index it from third crate.  when use it,  like a black box.

daogangtang:
or design a template tag to use it.

daogangtang:
such as {><}    :)

Keats:
So in the end it will go after #37 PR is in

@daogangtang Crate do not allow files to be package with so i'm not sure how that will work

@Alex-PK i'm thinking that only having kwargs might be more explicit in the end so i might go for that

Keats:
Closing in favour of filters in #44

PR #34, opened 2016-05-30
fix dir bug in windows

daogangtang:
fix dir bug in windows


Keats:
What's the bug? I thought Path was handling both windows and unix paths?

daogangtang:
hi, when find templates on windows, it will generate "views\\index.html" key and such as. so need to replace the key, to '/', to keep the same code as linux in business.

Keats:
I see, thanks.
Can you squash the commits and remove the commented out line please?

daogangtang:
ok, tonight do it.

PR #35, opened 2016-06-01
fix dir bug on windows

daogangtang:
Usually we render template like 
```
render("foo/bar.html", Context::new())
```
we use '/', but on windows, tera generate dir file system hash with format like "foo\\bar.html".

This will meet panic when execute the render command. This pr solve this.

### DIFF
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -36,6 +36,7 @@ impl Tera {
         // We are parsing all the templates on instantiation
         for entry in glob(dir).unwrap().filter_map(|e| e.ok()) {
             let path = entry.as_path();
+            println!("{:?}", path);
             // We only care about actual files
             if path.is_file() {
                 // We clean the filename by removing the dir given
@@ -49,6 +50,8 @@ impl Tera {
                 templates.insert(filepath.to_owned(), Template::new(&filepath, &input));
             }
         }
+
+        println!("templates {:?}", templates);
 
         Tera {
             templates: templates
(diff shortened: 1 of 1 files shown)


daogangtang:
how to rebase these submission?

src/tera.rs
```diff
@@ -49,13 +51,22 @@ impl Tera {
...
+        // println!("templates {:?}", templates);
+
         Tera {
             templates: templates
         }
     }
 
     pub fn render(&self, template_name: &str, data: Context) -> TeraResult<String> {
-        let template = self.templates.get(template_name).unwrap(); // TODO error handling
+        //let template = self.templates.get(template_name).unwrap(); // TODO error handling
```
Keats: you can remove that line as well

src/tera.rs
```diff
@@ -49,13 +51,22 @@ impl Tera {
...
+        let template = match self.templates.get(template_name) {
+            Some(tmpl) => tmpl,
+            None => {
+                println!("error in render {}", template_name);
```
Keats: and that one as well

Keats:
Do something like http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html for squashing commits

Keats:
Sorry, ended up pushing the fix myself (https://github.com/Keats/tera/commit/32126ba735d9c9357ae6f8be6e014acdf1867942)

Thanks for the report!

Keats:
@daogangtang can you try on master see if it works for you?

daogangtang:
all right now.

PR #37, opened 2016-06-06
Switch parser to pest

Keats:


Keats:
@dragostis started on it a bit

@Alex-PK you said you were interesting in parsing, so I guess that PR is relevant! Going to try to replace the handmade parser with https://github.com/dragostis/pest

Keats:
@dragostis how do i say "match everything until you hit X", for example parsing string can be `"hey_er/er.html"`?

src/parser2.rs
```diff
@@ -0,0 +1,51 @@
...
+
+        tag_start = { ["{%"] }
+        tag_end = { ["%}"] }
+
+        comment_start = { ["{#"] }
+        comment_end = { ["#}"] }
+
+        // math elements
+        decimals = @{ ["."] ~ ['0'..'9']+ }
+        int_ = @{ ["-"]? ~ (["0"] | ['1'..'9'] ~ ['0'..'9']*) }
```
dragostis: `int` and `float` are not keywords so they won't need underscores.

dragostis:
You can use: `!(something) ~ any`. `!` and `&` alone will not advance the position of the parser.

dragostis:
I would also recommend aligning the rules so that they're easier to read.

Keats:
I'll align everything once I stop moving everything around and add everything

I'm not sure I get what you mean by
> You can use: !(something) ~ any. ! and & alone will not advance the position of the parser.

Ie would matching a random string be that:
```rust
string = @{ ["\""] ~ (!(["\""]) ~ any )+ ~ ["\""]}
```
?

dragostis:
@Keats Precisely, only your `string` needs `*` instead of `+` to match the empty string as well. 😄

Keats:
Actually the rule `string = @{ ["\""] ~ (!(["\""]) ~ any )* ~ ["\""]}` causes a 
`<pest macros>:141:1: 148:36 error: recursion limit reached while expanding the macro grammar`

dragostis:
Yes. I should mention it in the docs. You need to increase the recursion [limit](https://github.com/dragostis/pest/blob/master/tests/json.rs#L8).

Keats:
Hmm even with `#![recursion_limit = "80"]` it's failing, why would it recurse so much in the first place?

dragostis:
You probably need to find a higher value. For example, the Lua example in pest needs about 400 to compile. The problem is that you need a lot of recursive macro calls in order to construct the parser properly.

src/parser2.rs
```diff
@@ -0,0 +1,86 @@
...
+        content = {
+            text |
+            comment_tag |
+            block_tag ~ content ~ endblock_tag |
+            if_tag ~ content ~ (elif_tag ~ content)* ~ (else_tag ~ content)? ~ endif_tag |
+            for_tag ~ content ~ endfor_tag
+        }
+
+        // top level node
+        template = { extends_tag? ~ content ~ eoi }
```
Keats: @dragostis I'm not sure how to represent the template as a whole, the test below gives a stack overflow which is due to my lack of understanding in that I think. 
Not sure if it's in the scope of pest or PEG in general but a helper for something like `take_until` would be nice

src/parser2.rs
```diff
@@ -0,0 +1,86 @@
...
+
+#[cfg(test)]
+mod tests {
+    use pest::prelude::*;
+    use super::Rdp;
+
+    #[test]
+    fn test_only_text() {
+        let mut parser = Rdp::new(StringInput::new("Hello\n 世界"));
+        assert!(parser.expression());
```
Keats: @dragostis i can't find docs for those methods so i'm not really what this is doing, `.expression()` doesn't tell me much, is it doing the parsing?

src/parser2.rs
dragostis: PEGs don't support [left recursion](https://en.wikipedia.org/wiki/Parsing_expression_grammar#Indirect_left_recursion) which is probably the root cause of the stack overflow. I'm on the train right now, but I'll have a look later today or tomorrow.

src/parser2.rs
dragostis: The `grammar!` macro creates methods with the names of the rules on the parser instance.

src/parser2.rs
Keats: It might be just malformed expression from my end, I'm busy today so I will have another look tomorrow

src/parser2.rs
Keats: ah! definitely not what i wanted then

dragostis: Would this be an issue? You can encapsulate the Parser inside another structure and just call your main rule from a parse method.

Keats: nop it makes perfect sense, It's mainly that in the calculator example you are using `.expression()` everywhere so I thought it was something special

dragostis: You will probably want to use `template` there. 😄 

The way I constructed the JSON and Lua parsing examples was with a TDD, bottom-up approach. I'd write small tests for simpler rules first (e.g. does it match every type of int, float, string?) and then continue like this until reaching the top-level rule. This might work well here too.

Keats: yep, i'll probably do that, thanks

Keats: I'd recommend to update your examples/doc to use

```rust
assert_eq!(parser.queue(), &queue);
```

rather than
```rust
assert!(parser.queue().iter().eq(&queue));
```

dragostis: Good idea! Thanks. 👍

Keats:
Started adding tests, works great for. 
Next step is figuring out how to make the `expression` rule with the precedence

dragostis:
[This](https://github.com/dragostis/pest/blob/master/tests/grammar.rs#L316) is how precedence climbing works in detail.

I'm about to release `0.2.0` right now.

Keats:
Should I separate each op kind (plus, minus, equal, etc) in their own token or is the current way ok and i can retrieve the sign later on?

dragostis:
@Keats Either way works, but by having separate rules, you will be able to compare `Token`'s `Rule` instead of a captured `&str`. Matching `Rule` should be faster and more efficient than a `&str`. In the calculator example I matched `Rule`s.

src/parser2.rs
```diff
@@ -0,0 +1,237 @@
...
+            Token { rule: Rule::op_plus, start: 6, end: 7 },
+            Token { rule: Rule::int, start: 8, end: 9 },
+            Token { rule: Rule::op_gt, start: 10, end: 11 },
+            Token { rule: Rule::int, start: 12, end: 13 }
+        ]);
+    }
+
+    #[test]
+    fn test_expression_identifier_logic_complex() {
+        let mut parser = Rdp::new(StringInput::new("1 > 2 || 3 == 4 && admin"));
```
Keats: precedence is working for all the tests above but fail at this one.

I think i need to add `expression` in the precedence climbing but then it stackoverfows

dragostis: This fails because there is no `||` defined in the grammar. 😛

Keats: well that's a stupid mistake haha, forgot I'm switching from `|| &&` to `or and` >_<, it works now

Does that mean it will panic if it finds an unexpected char instead of erroring? Ideally it should have an error token that I can improve later on (ie, you used `||` but Tera uses `or` so replace `||` by `or` on line 23, col 19)

dragostis: It doesn't panic. If you run `expression()` it will leave some unmatched `Rule`s inside `Parser::expected()`.

If you want Tera to know about `||`, you can add a rule `op_wrong_or` that you can pattern-match for in `process!`. Unfortunately, pattern matching inside an infix operation is not going to be very pretty, but it's definitely serviceable.

It would be easier to add these exception to the `grammar!`, but I'm planning to keep it clean, without any code.

dragostis: What I mean is that you can just add `op_wrong_or ` to the same precedence and when you're matching the `Token`, you can have it `println!("|| was deprecated; use or")` or `panic!("use or instead of ||").`

Keats:
@dragostis thanks, only need to fix the `content` one that is failing for some reason (`thread 'parser2::tests::test_content' panicked at 'assertion failed: parser.end()', src/parser2.rs:99`) and it seems that the lexing part will be done after that

src/parser2.rs
```diff
@@ -0,0 +1,310 @@
...
+        block_tag    = { tag_start ~ ["block"] ~ identifier ~ tag_end }
+        if_tag       = { tag_start ~ ["if"] ~ expression ~ tag_end }
+        elif_tag     = { tag_start ~ ["elif"] ~ expression ~ tag_end }
+        else_tag     = { tag_start ~ ["else"] ~ tag_end }
+        for_tag      = { tag_start ~ ["for"] ~ identifier ~ ["in"] ~ identifier ~ tag_end }
+        endblock_tag = { tag_start ~ ["endblock"] ~ identifier ~ tag_end }
+        endif_tag    = { tag_start ~ ["endif"] ~ tag_end }
+        endfor_tag   = { tag_start ~ ["endfor"] ~ tag_end }
+
+        text = { (!(block_start) ~ any )* }
```
dragostis: `content` matches if `text` matches. And `text` matches for 0 or more `any` which *always* matches. It will just match the empty string `""`. This may be a hint that something is not good.

dragostis: Maybe you should add `text` last?

Keats: hmm if i put text last, I still get the error on my end:
```rust
        text = { (!(block_start) ~ any )* }

        content = {
            variable_tag |
            comment_tag |
            block_tag ~ content ~ endblock_tag |
            if_tag ~ content ~ (elif_tag ~ content)* ~ (else_tag ~ content)? ~ endif_tag |
            for_tag ~ content ~ endfor_tag |
            text
        }
```

But text should match for any until we get a `block_start` no?

dragostis: What error do you get? If you put `text` last, then it makes sense to match it like that. But shouldn't it also test for `{%` along with `{{`?

Keats: `block_start` is `block_start    = _{ variable_start | tag_start | comment_start }`

I get `	thread 'parser2::tests::test_content' panicked at 'assertion failed: parser.end()', src/parser2.rs:98`

dragostis: `end()` fails because `whitespace` gets run only in front of any rule. (including `eoi`) This example has some white-space at the end which is why `end()` returns `false`.

Keats: got it thanks, only the template lexing left to debug!

src/parser2.rs
```diff
@@ -0,0 +1,312 @@
...
+    #[test]
+    fn test_template() {
+        let mut parser = Rdp::new(StringInput::new("
+            {# Greeter template #}
+            Hello {% if i18n %}世界{% else %}world{% endif %}
+            {% for country in countries %}
+                {{ loop.index }}.{{ country }}
+            {% endfor %}"));
+        assert!(parser.template());
+        expect_tokens(&parser, vec![
```
Keats: so this one seems stuck in an infinite loop somewhere @dragostis 

Looking at template `template = _{ extends_tag? ~ content* ~ eoi }`, it seems like it should work. Is there a way to debug those?

dragostis: This is generally one of the least coolest things to debug, but it shouldn't be an issue with a debugger. Let me take a look.

dragostis: One little issue is that you probably shouldn't have void `text` in a `*`, because it will match infinitely. `text` should probably be at least one character long, so it should be `text = { (!block_start ~ any )+ }`. This doesn't solve the whole issue, though.

Keats: I'm not sure if that makes sense but maybe there could be a `impl_rdp_debug!` that would print out what the parser is doing

dragostis: I've thought about that, but I'd be in favour of just using a visual debugger right now. I think it's pretty easy to debug with `expected()` and `queue()`.

If you add `+` to `text`, you need to make `content` optional in all the blocks where it is used with a `?`. I did this, but there's still some little issue left.

dragostis: Now it fails because you have more than 1 `content` inside the for clause. So:

* `text` should end with a `+`
* `content` should **probably** end with a `*`

dragostis: Here is what I've changed:

```
text = { (!(block_start) ~ any )+ }
content = _{
    variable_tag |
    comment_tag |
    block_tag ~ content* ~ endblock_tag |
    if_tag ~ content* ~ (elif_tag ~ content?)* ~ (else_tag ~ content*)? ~ endif_tag |
    for_tag ~ content* ~ endfor_tag |
    text
}
```

I would also recommend using some kind of rule to enclose `content*` inside of, because matching `content*` inside the current `process!` macro might be hard.

Keats: Adding a "how to debug" section on the doc would be helpful!

The `if_tag` seems incorrect, `(elif_tag ~ content?)*` -> `(elif_tag ~ content*)*` but I have the `assertion failed: parser.end()` error now. Could it be due to `eoi`?

dragostis: This appears to be the first bug in the grammar! 🎆  Lemme see what's causing it.

Keats: o/

I'm probably done for today, next step after the lexer is making the AST which *should* be easier.
Thanks a lot for the help so far!

dragostis: Solved the issue. It was very much a corner case. The test should pass now.

Keats:
Great @dragostis !

AST building tomorrow if i can find some time

dragostis:
Right now it's a bit hard to process `Token`s coming from `+` or `*, so I'm adding an extra matcher in `process!` to make it much easier.

Keats:
I will start the AST this afternoon or this weekend depending how my work is going.
Maybe opening a gitter channel would be easier to communicate than this PR

dragostis:
@Keats I've just added you on Gitter.

Keats:
I'm not entirely sure how to get started with the `process!` so i'll catch you on gitter this weekend/next week

src/parser2.rs
```diff
@@ -0,0 +1,330 @@
...
+                ["0"] ~ ["."] ~ ['0'..'9']+ |
+                ['1'..'9'] ~ ['0'..'9']* ~ ["."] ~ ['0'..'9']+
+        }
+
+        identifier = @{
+            (['a'..'z'] | ['A'..'Z'] | ["_"]) ~
+            (['a'..'z'] | ['A'..'Z'] | ["_"] | ["."] | ['0'..'9'])*
+        }
+        // matches anything between 2 double quotes
+        string = @{ ["\""] ~ (!(["\""]) ~ any )* ~ ["\""]}
```
sunng87: How about escaped quot like `"hello\"world"`?

sunng87: I'm using rule `{ ["\""] ~ (!["\""] ~ (["\\\""] | any))* ~ ["\""] }` for escaped `"` but it doesn't work. @dragostis any ideas?

dragostis: @sunng87 The [JSON grammar](https://github.com/dragostis/pest/blob/master/tests/json.rs#L26) provides an example on how to implement escapes.

Keats: I have to say I didn't really bother with edge cases like that yet

sunng87: Thank you. It was a mistake in my test case. The grammar works.

Keats:
@dragostis seems like the `process!`  roughly work now (probably still some issues with my custom whitespace handling code but that's not pest).

I'd appreciate a review of `grammar!` and `process!` when you have time!

dragostis:
@Keats I'll take a look sometime next week. Most probably Monday. 😃

Keats:
No worries and thanks!

Have to rewrite the rendering now

Keats:
It's not possible to pass some arbitrary data to process right? If i wanted to panic on some incorrect matching, I wouldn't be able to pass the filename/line/col so it would be better for the `process!` fn to return `Result` right?

dragostis:
Definitely. I think it is a far more elegant solution to use `Result`. 😃

Keats:
Wired in the pest parser (sorry the diff is trying to be too smart there so it's kinda hard to read the parser.rs diff).

This branch:
```
test bench_parsing   ... bench:      10,439 ns/iter (+/- 3,785)
test bench_rendering ... bench:       4,732 ns/iter (+/- 1,923)
```

Master:
```
test bench_parsing   ... bench:      29,290 ns/iter (+/- 19,324)
test bench_rendering ... bench:       3,862 ns/iter (+/- 702)
```

And all with -1000 loc, pretty sweet win!
Looking forward to the whitespace improvements

Keats:
@dragostis merging that for now, will do the whitespace with pest 0.4 in another pr

PR #47, opened 2016-08-31
Implement lower filter

orhanbalci:
Hi. I saw your project on rust newsletter and wanted to contribute.


Keats:
Thanks for that!

You should add *.swp in your global gitignore so these files are automatically ignored

orhanbalci:
Thanks for your comment

PR #49, opened 2016-09-01
Added Filters

Peternator7:
Implemented Several of the filters requested on this-week-in-rust. First, Last, and Join for the Array filters. Rust fmt also added a comma here and there.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -45,9 +45,5 @@ quick_error! {
             display("Filter `{}` received an incorrect type for arg `{}`: got {:?} but expected a {}", filter_name, arg_name, arg_value, expected_type)
             description("incorrect filter arg type")
         }
-        EmptyArray(filter_name: String) {
-            display("Array provided is empty, but filter `{}` is trying to access an element", filter_name)
-            description("Tried retriving an element in an empty array.")
-        }
     }
 }
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -46,7 +46,7 @@ quick_error! {
             description("incorrect filter arg type")
         }
         EmptyArray(filter_name: String) {
-            display("Array provided is empty, but filter {} is trying to access an element", filter_name)
+            display("Array provided is empty, but filter `{}` is trying to access an element", filter_name)
             description("Tried retriving an element in an empty array.")
         }
     }
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -28,7 +28,7 @@ quick_error! {
             description("filter not found")
         }
 
-        // Runtime errors
+// Runtime errors
         NotANumber(name: String) {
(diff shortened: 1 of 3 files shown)


Keats:
Thanks for that!
I think `first` and `last` should not be errors as it seems common enough that it can happen and that isn't necessarily an error. Returning `""` should be enough and maybe `None` later on when Tera will have its own serialization.

src/filters/array.rs
```diff
@@ -0,0 +1,117 @@
...
+    if let Some(val) = arr.last() {
+        Ok(val.clone())
+    } else {
+        Err(TeraError::EmptyArray("last".to_string()))
+    }
+}
+
+pub fn join(value: Value, args: HashMap<String, Value>) -> TeraResult<Value> {
+    let arr = try_get_value!("join","value",Vec<Value>, value);
+    let separator = match args.get("d") {
```
Keats: can you name that arg `sep`?

src/filters/array.rs
```diff
@@ -0,0 +1,117 @@
...
+        Some(val) => try_get_value!("truncate", "d", String, val.clone()),
+        None => "".to_string(),
+    };
+    let mut reduction = arr.iter()
+        .fold("".to_string(),
+              |acc, val| acc + &*separator + &*val.render());
+    if reduction.len() > 0 {
+        // if we had any elements, trim the separator off the front.
+        reduction = reduction.split_at(separator.len()).1.to_string();
+    }
```
Keats: `Vec` have a join method https://doc.rust-lang.org/std/slice/trait.SliceConcatExt.html#tymethod.join, is it possible to use it?

src/filters/array.rs
```diff
@@ -0,0 +1,117 @@
...
+    } else {
+        Err(TeraError::EmptyArray("last".to_string()))
+    }
+}
+
+pub fn join(value: Value, args: HashMap<String, Value>) -> TeraResult<Value> {
+    let arr = try_get_value!("join","value",Vec<Value>, value);
+    let separator = match args.get("d") {
+        Some(val) => try_get_value!("truncate", "d", String, val.clone()),
+        None => "".to_string(),
```
Keats: default to `" "` would be better imo

Keats: maybe not actually, leave it like that for now

Peternator7: Yeah, I thought about that. The docs default to "" though so I thought at least initially consistency might be best.

src/errors.rs
```diff
@@ -28,7 +28,7 @@ quick_error! {
             description("filter not found")
         }
 
-        // Runtime errors
+// Runtime errors
```
Keats: can you align that again?

Keats:
Thanks for that!

PR #50, opened 2016-09-01
Replace filter implemented

orhanbalci:


Keats:
ah sorry  I missed the `count` argument from Jinja2.
I think we can drop that and just replace everything using https://doc.rust-lang.org/std/string/struct.String.html#examples-61

Sorry for the extra work

Keats:
also opens the discussion of mandatory args in filters.
`old` and `new` are required i'd say, defaulting to `""` is a bit weird.

Tera should probably return an error if one of them is not there but i'll change that after this PR is merged

orhanbalci:
I can add FilterRequiredArgNotFound to TeraError and return that if you prefer

src/filters/string.rs
```diff
@@ -116,4 +135,34 @@ mod tests {
         assert!(result.is_ok());
         assert_eq!(result.unwrap(), to_value(&4));
     }
+
+    #[test]
+    fn test_replace() {
```
Keats: you only need that test since we are using the built-in `replace`, you can remove the 2 other tests

orhanbalci: removed unnecessary tests also added capitalize filter

Keats:
Thanks!

PR #52, opened 2016-09-02
Addslashes filter implemented

orhanbalci:


nabijaczleweli:
oh god, so many merges

Keats:
Can you add the testcases from https://github.com/django/django/blob/ecb59cc6579402b68ddfd4499bf30edacf5963be/tests/template_tests/filter_tests/test_addslashes.py as well @orhanbalci please?

@nabijaczleweli github has squash and merge no so that's not a big issue but I'd recommend @orhanbalci to `git reset --hard` on my master and cherrypick `6529c98` on top, easier than merging!

orhanbalci:
@Keats should we also escape backslashes?

Keats:
I'd say yes if that's what django is doing

orhanbalci:
@Keats  testcases added

Keats:
Thanks!

PR #53, opened 2016-09-02
Filters Length Reverse

Peternator7:
Implemented filters for length and reverse. For length, the django docs says it returns 0 for an undefined obj so I went ahead and applied that logic to any non-string/array type.

### DIFF
--- a/src/filters/common.rs
+++ b/src/filters/common.rs
@@ -4,6 +4,8 @@ use std::collections::HashMap;
 use serde_json::value::{Value, to_value};
 use errors::{TeraError, TeraResult};
 
+// Returns the number of elements in an array or the number of
+// characters in a string.
 pub fn length(value: Value, _: HashMap<String, Value>) -> TeraResult<Value> {
     match value {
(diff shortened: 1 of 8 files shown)


src/filters/common.rs
```diff
@@ -0,0 +1,98 @@
+/// Filters operating on string
```
Keats: on string -> on several types

src/filters/common.rs
```diff
@@ -0,0 +1,98 @@
...
+pub fn reverse(value: Value, _: HashMap<String, Value>) -> TeraResult<Value> {
+    match value {
+        Value::Array(arr) => {
+            // Clone the array so that we don't mutate the original.
+            let mut rev = arr.clone();
+            rev.reverse();
+            Ok(to_value(&rev))
+        }
+        Value::String(s) => {
+            use std::iter::FromIterator;
```
Keats: put that import at the top of the file

Keats:
Looks good after my 2 comments!

Keats:
Thanks!

PR #54, opened 2016-09-07
title filter added

orhanbalci:
implemented title filter

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ clippy = {version = "~0.0.77", optional = true}
 pest = "0.4"
 quick-error = "1.1.0"
 slug = "0.1"
-regex = "0.1.71"
+regex = "0.1"
 
 [features]
 default = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ serde_json = "0.8"
 clippy = {version = "~0.0.77", optional = true}
 pest = "0.4"
 quick-error = "1.1.0"
+slug = "0.1"
 
 [features]
 default = []
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ clippy = {version = "~0.0.77", optional = true}
 pest = "0.4"
 quick-error = "1.1.0"
 slug = "0.1"
+regex = "0.1.71"
 
 [features]
 default = []
--- a/src/filters/string.rs
+++ b/src/filters/string.rs
@@ -337,7 +337,6 @@ mod tests {
             ("b7>b2!", "b7>b2!"),
             ("</fe", "</fe"),
             ("<x>b<y>", "b"),
-            //(r#"a<p onclick="alert('<test>')">b</p>c"#, "abc"),
             (r#"a<p a >b</p>c"#, "abc"),
             (r#"d<a:b c:d>e</p>f"#, "def"),
             (r#"<strong>foo</strong><a href="http://example.com">bar</a>"#, "foobar"),
--- a/src/filters/string.rs
+++ b/src/filters/string.rs
@@ -159,7 +159,7 @@ pub fn title(value: Value, _: HashMap<String, Value>) -> TeraResult<Value> {
 pub fn striptags(value : Value, _: HashMap<String, Value>) -> TeraResult<Value> {
     let s = try_get_value!("striptags", "value", String, value);
     let tag_pattern = Regex::new(r"(<!--.*?-->|<[^>]*>)").unwrap();
-    Ok(to_value(&tag_pattern.split(&s).map(|x| x.replace("\\n","").trim().to_string()).filter(|x| !x.is_empty()).collect::<Vec<String>>().join(" ")))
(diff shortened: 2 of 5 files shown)


Keats:
Can you add the tests from jinja linked in the other issue? The closer we are to django/jinja the better

orhanbalci:
Jinja implementation respects whitespaces whereas mine does not. Other test cases will fail if I add them.  If we want to respect whitespaces I should update implementation. Jinja uses regexp to define splitting patterns. I can apply the same strategy if you want.

Keats:
Yeah I think it should respect whitespace, feel free to pull the regex crate, we might need it for other stuff like `striptag` anyway

orhanbalci:
@Keats  rewrote title filter using regex

Cargo.toml
```diff
@@ -16,6 +16,7 @@ serde_json = "0.8"
 clippy = {version = "~0.0.77", optional = true}
 pest = "0.4"
 quick-error = "1.1.0"
+regex = "0.1.71"
```
Keats: just put `regex = "0.1"` so we get the updates as well

src/filters/string.rs
```diff
@@ -93,6 +94,62 @@ pub fn addslashes(value: Value, _: HashMap<String, Value>) -> TeraResult<Value>
 
 }
 
+/// Capitalizes each word in the string
+pub fn title(value: Value, _: HashMap<String, Value>) -> TeraResult<Value> {
+    let s = try_get_value!("title", "value", String, value);
+    let split_pattern = Regex::new(r"([-\s\(\{\[<]+)").unwrap();
```
Keats: we should be able to compile that regex into static but that should be ok for now

orhanbalci:
I also added striptags filter

src/filters/string.rs
```diff
@@ -93,6 +94,69 @@ pub fn addslashes(value: Value, _: HashMap<String, Value>) -> TeraResult<Value>
...
+            Ok(to_value(&result))
+        }
+    }
+}
+
+///Removes html tags from string
+pub fn striptags(value : Value, _: HashMap<String, Value>) -> TeraResult<Value> {
+    let s = try_get_value!("striptags", "value", String, value);
+    let tag_pattern = Regex::new(r"(<!--.*?-->|<[^>]*>)").unwrap();
+    Ok(to_value(&tag_pattern.split(&s).map(|x| x.replace("\\n","").trim().to_string()).filter(|x| !x.is_empty()).collect::<Vec<String>>().join(" ")))
```
Keats: can we use the regex replace method instead?

src/filters/string.rs
```diff
@@ -209,4 +273,44 @@ mod tests {
...
+            assert_eq!(result.unwrap(), to_value(expected));
+        }
+    }
+
+    #[test]
+    fn test_striptags() {
+        let tests = vec![
+            (r"<b>Joel</b> <button>is</button> a <span>slug</span>", "Joel is a slug"),
+            (r#"<p>just a small   \n <a href="x"> example</a> link</p>\n<p>to a webpage</p><!-- <p>and some commented stuff</p> -->"#,
+            "just a small example link to a webpage"),
```
Keats: if you can add the tests from https://github.com/django/django/blob/ecb59cc6579402b68ddfd4499bf30edacf5963be/tests/utils_tests/test_html.py#L68-L88, it would be nice

src/filters/string.rs
Keats: and why is there special handling for `\n`?

orhanbalci: In django test case it removed \n. Rusts trim method does not remove \n

Keats: using django 1.10:

```
In [5]: strip_tags("<p> blalb \n erer</p>")
Out[5]: ' blalb \n erer'
```
That might just be one of their test quirk, you can just remove that special handling

Keats:
@orhanbalci do you want me to finish that PR?

orhanbalci:
@Keats I was on vacation and offline for a while sorry for the  delay. I'll try to finish this up

orhanbalci:
@Keats  using regular expression from jinja  https://github.com/pallets/jinja/blob/master/jinja2/utils.py#L27 but fails for the test case https://github.com/django/django/blob/ecb59cc6579402b68ddfd4499bf30edacf5963be/tests/utils_tests/test_html.py#L82

src/filters/string.rs
```diff
@@ -209,4 +273,58 @@ mod tests {
...
+            (r"<adf>a", "a"),
+            (r"</adf>a", "a"),
+            (r"<asdf><asdf>e", "e"),
+            (r"hi, <f x", "hi, <f x"),
+            ("234<235, right?", "234<235, right?"),
+            ("a4<a5 right?", "a4<a5 right?"),
+            ("b7>b2!", "b7>b2!"),
+            ("</fe", "</fe"),
+            ("<x>b<y>", "b"),
+            //(r#"a<p onclick="alert('<test>')">b</p>c"#, "abc"),
```
Keats: remove that line

Keats:
Seems good then, can you add the doc in readme.md for those 2 filters and rebase?

Keats:
Thank you!

PR #55, opened 2016-09-12
Add urlencode filter

foophoof:
Not sure if you're okay with a new dependency just for this.

I implemented most of the tests from https://github.com/django/django/blob/1ec1633cb294d8ce2a65ece6b56c258483596fba/tests/template_tests/filter_tests/test_urlencode.py, but `'fran\xe7ois & jill'` is not valid UTF-8, so Rust doesn't let me make that into a string literal.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ clippy = {version = "~0.0.77", optional = true}
 pest = "0.4"
 quick-error = "1.1.0"
 slug = "0.1"
+url = "1.2.0"
 
 [features]
 default = []
--- a/README.md
+++ b/README.md
@@ -246,3 +246,11 @@ Returns the length of an array or a string, 0 if the value is not an array.
 
 #### reverse
 Returns a reversed string or array
+
+#### urlencode
+Percent-encodes a string.
+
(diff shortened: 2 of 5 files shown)


Keats:
Ha sorry I never noticed that the django `urlencode` had a parameter. Have you used it before? Even in django own source code, it seems only used once and only for `/`.  
I'm thinking something with no args like Jinja might be better but I just don't know how common it is to use the safe arg in Django.

Adding the dependency to `url` is fine

foophoof:
I have not used this before, I mostly just saw the request for contributions in TwiR a while back and looked for filters I could implement. Most URL encoding functions I see have some way to pass in which characters to quote (or not to quote), so I'm assuming they are used often enough?

Keats:
Looks good to me then, can you just add a quick note about it in the filters section in the README please?

Keats:
@foophoof ping, I can finish the PR if you don't have time

foophoof:
@Keats Sorry about the delay, added a blurb to the README and fixed the conflicts against master.

Keats:
Thanks!

PR #58, opened 2016-09-20
Implement testers

### LINKED ISSUE #56: Add support for variable tests
Hello!

Great project you have going on here!

One feature that I feel is sorely missing is support for [tests](http://jinja.pocoo.org/docs/dev/templates/#tests). Have you considered implementing this? The case when I most want to use this is to check if a variable has been defined. For example, I'd like to be able to write:

```
{% if some_var is defined %}
  {{ some_var }}
{% endif %}
```

It would also be nice to make this pluggable to allow anyone to implement a test.

SergioBenitez:
This PR contributes an implementation of [tests](http://jinja.pocoo.org/docs/dev/templates/#tests) for Tera. Closes #56.

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -18,11 +18,14 @@ pub enum Node {
     Logic {lhs: Box<Node>, rhs: Box<Node>, operator: String},
 
     If {condition_nodes: LinkedList<Node>, else_node: Option<Box<Node>>},
-    // represents a if/elif block and its body (body is a List)
+    // represents a if/elif block. condition (Math, Logic, Test), body (a List)
     Conditional {condition: Box<Node>, body: Box<Node>},
 
     For {variable: String, array: String, body: Box<Node>},
     Block {name: String, body: Box<Node>},
+
+    // params are expressions
+    Test {expression: Box<Node>, name: String, params: LinkedList<Node>},
 
     // For now all params are strings. We still need to differentiate between
     // user input (ie "dd/mm/YY" arg in a date filter)
@@ -84,6 +87,7 @@ impl_rdp! {
             (['a'..'z'] | ['A'..'Z'] | ["_"]) ~
             (['a'..'z'] | ['A'..'Z'] | ["_"] | ['0'..'9'])*
         }
+
         // named args
         fn_arg_value = @{ boolean | int | float | string | identifier }
         fn_arg       = @{ simple_ident ~ ["="] ~ fn_arg_value }
@@ -98,6 +102,17 @@ impl_rdp! {
         }
         identifier_with_filter = { identifier ~ filters }
         idents = _{ identifier_with_filter | identifier }
+
+        // Variable tests.
+        test_fn_param = { expression }
+        test_fn_params = {
+            test_fn_param
+            | (["("] ~ test_fn_param ~ ([","] ~ test_fn_param)* ~ [")"])
+        }
+        test_fn = !@{ simple_ident ~ test_fn_params? }
+        // explicit whitespace is needed so that `if is_defined` doesn't get
+        // parsed as a test.
+        test = { ["is"] ~ test_fn }
 
         // Precedence climbing
         expression = _{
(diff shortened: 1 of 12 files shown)


src/render.rs
```diff
@@ -197,95 +240,35 @@ impl<'a> Renderer<'a> {
...
+                            rhs_val = Value::F64(rhs_val.as_f64().unwrap());
                         }
+
+                        let result = match operator.as_str() {
+                            "==" => lhs_val == rhs_val,
+                            "!=" => lhs_val != rhs_val,
+                            _ => unreachable!()
+                        };
+
+                        Ok(result)
```
Keats: :heart:

src/render.rs
```diff
@@ -197,95 +240,35 @@ impl<'a> Renderer<'a> {
...
+                        // Monomorphize number vals.
+                        if lhs_val.is_number() || rhs_val.is_number() {
+                            if !lhs_val.is_number() || !rhs_val.is_number() {
+                                return Ok(false);
+                            }
+
+                            // Since Tera only support 32 bit integers, this
+                            // actually preserves all of the precision. If Tera
+                            // switches to 64-bit values, use std::f32::EPSILON
+                            // to get an approximation as before.
```
Keats: that's something I was thinking about actually, I'd be tempted to always use 64-bit

Keats:
As mentioned in the issue, I think I'd rather start with just overloading `if` but this PR is very good.
If you can keep this PR as is and open a new one for the if overloading, I'll use that one as a base to implement tests if we end up adding them.

I love the changes to `render.rs` and will do them myself later if you don't want to redo them in another PR

SergioBenitez:
I'd like to argue in favor of accepting this PR, and testers, now.

While it does seem that some projects abuse filters as testers, it should be noted that this is exactly that: an _abuse_ of filters. Filters and testers have two, orthogonal purposes. A filter is meant to convert one value into another, almost exclusively of the same type. A tester, on the other hand, checks that some value meets some condition. These are two different things - two different concepts - and they deserve two different syntaxes with different semantics and expectations. The right tool should be used for the job.

Furthermore, I feel that as a templating library, you should not restrict programmers from programming in the way they see fit. Personally, I'd like to expose testers libraries that aren't filters. For example, I have an idea for a tester that checks a value against the internal state of the application. The only reasonable thing to use here is a tester. A filter is the _wrong_ thing to use. It would be a shame if the templating library forces me to use the wrong thing, especially when there is an implementation of the _right_ thing.

Anyway, I don't see the harm in accepting this PR. People use testers in Jinja (see StackOverflow), and I would imagine that the Rust crowd will want to use the right tool for the job.

Keats:
ok you've convinced me, let me do the code review

src/parser.rs
```diff
@@ -867,6 +1036,38 @@ mod tests {
...
+        assert!(parser.end());
+    }
+
+    #[test]
+    fn test_unary_test() {
+        let mut parser = Rdp::new(StringInput::new("is equalto other"));
+        assert!(parser.test());
+        assert!(parser.end());
+
+        let mut parser = Rdp::new(StringInput::new("is equalto(other)"));
```
Keats: can we have only one syntax or is there a reason 2 are needed?

src/render.rs
```diff
@@ -167,22 +166,64 @@ impl<'a> Renderer<'a> {
...
+                Ok(Value::F64(val as f64))
+            },
+            Bool(b) => {
+                Ok(Value::Bool(b))
+            },
+            _ => unreachable!()
+        }
+    }
+
+    // TODO: clean up this, too ugly right now for the == and != nodes
```
Keats: i think you can remove that TODO now

src/render.rs
```diff
@@ -167,22 +166,64 @@ impl<'a> Renderer<'a> {
...
+        match node {
+            Identifier { .. } => {
+                Ok(self.eval_ident(&node).map(|v| v.is_truthy()).unwrap_or(false))
+            },
+            Test { expression, name, params } => {
+                let tester = try!(self.tera.get_tester(&name));
+                let context = match self.context.as_object() {
+                    Some(map) => map,
+                    None => return Err(Internal("context is not an object.".into()))
+                };
```
Keats: How can that happen? you need to pass a context to render so you should always have an object

Keats:
Can you also add a small note on the README about the behaviour of if with undefined values? thanks!

src/parser.rs
SergioBenitez: This is what Jinja2 does. It's pretty neat to be able to write "my_var is odd" instead of "my_var is odd()".

src/render.rs
SergioBenitez: Done.

src/render.rs
SergioBenitez: It's just there for sanity checking and defensive programming. For example, if `context` changes in the future to allow things that aren't maps, this will catch that with a helpful error message.

SergioBenitez:
Note added.

src/render.rs
Keats: I think you can safely remove that, the next time the context might change is when moving from JSON to something specific to Tera and those lines will not be relevant then

You can safely unwrap

SergioBenitez:
@Keats Everything good?

Keats:
Yep thanks!

PR #61, opened 2016-10-03
Support indexing

Keats:


src/render.rs
```diff
@@ -70,7 +70,7 @@ impl<'a> Renderer<'a> {
     fn lookup_variable(&self, key: &str) -> TeraResult<Value> {
         // Look in the plain context if we aren't in a for loop
         if self.for_loops.is_empty() {
-            return self.context.lookup(key).cloned()
+            return self.context.pointer(&("/".to_string() + &key.replace(".", "/"))).cloned()
```
SergioBenitez: I think this would look nicer, and avoid an allocation, if split into:

```rust
let path = &["/", &key.replace(".", "/")].join("");
return self.context.pointer(path).cloned()
    .ok_or_else(|| FieldNotFound(key.to_string()));
```

src/render.rs
```diff
@@ -102,7 +102,8 @@ impl<'a> Renderer<'a> {
 
         // dummy statement to satisfy the compiler
         // TODO: make it so that's not needed
-        self.context.lookup(key).cloned().ok_or_else(|| FieldNotFound(key.to_string()))
+        self.context.pointer(&("/".to_string() + &key.replace(".", "/"))).cloned()
```
Keats: Slice patterns don't seem stable yet and it still seems a bit uncertain looking at https://github.com/rust-lang/rust/issues/23121

PR #63, opened 2016-10-04
Add tests and remove using AST in tests fn

Keats:


Keats:
@SergioBenitez remove the AST from the tests fn, error reporting still kinda suck as it doesn't tell you files/lines but that's the same issue with filters and I need to come up with a better way for both

README.md
```diff
@@ -215,12 +215,34 @@ is defined, you would write:
 ... don't use user here ...
 {% end %}
 ```
+Note that testers allow expressions so the following is a valid test as well:
```
SergioBenitez: Needs comma after "expressions": "expressions, so".

src/render.rs
```diff
@@ -91,7 +84,7 @@ impl<'a> Renderer<'a> {
                 // might be a struct or some nested structure
                 if key.contains('.') {
                     let new_key = key.split_terminator('.').skip(1).collect::<Vec<&str>>().join(".");
-                    return value.pointer(&self.get_json_pointer(&new_key)).cloned().ok_or_else(|| FieldNotFound(key.to_string()));
+                    return value.pointer(&get_json_pointer(&new_key)).cloned().ok_or_else(|| FieldNotFound(key.to_string()));
```
SergioBenitez: This line is _really_ long. Either split it up, or use the idiomatic Rust style for cases like this:

```rust
value.pointer(&get_json_pointer(&new_key))
     .cloned()
     .ok_or_else(|| FieldNotFound(key.to_string()));
```

src/render.rs
```diff
@@ -108,8 +101,7 @@ impl<'a> Renderer<'a> {
 
         // dummy statement to satisfy the compiler
         // TODO: make it so that's not needed
-        self.context.pointer(&self.get_json_pointer(key)).cloned()
-            .ok_or_else(|| FieldNotFound(key.to_string()))
+        self.context.pointer(&get_json_pointer(key)).cloned().ok_or_else(|| FieldNotFound(key.to_string()))
```
SergioBenitez: Same thing here, though it's not as bad.

Also, why not use `unreachable!` here?

src/render.rs
```diff
@@ -207,17 +199,11 @@ impl<'a> Renderer<'a> {
...
-                let context = match self.context.as_object() {
-                    Some(map) => map,
-                    None => return Err(Internal("context is not an object.".into()))
-                };
-
-                let mut value_params = LinkedList::new();
+                let mut value_params = vec![];
                 for param in params {
-                    value_params.push_back(try!(self.eval_expression(param)));
+                    value_params.push(try!(self.eval_expression(param)));
```
SergioBenitez: Why not:

```rust
let value_params: Vec<_> = params.map(|p| try!(self.eval_expression(p)).collect()l
```

src/testers.rs
```diff
@@ -1,30 +1,123 @@
...
-        -> TeraResult<bool> {
-    if params.len() != 0 {
-        return Err(TeraError::TestError("defined".to_string(),
-            "defined should not be called with parameters".to_string()))
+use serde_json::value::{Value};
+use context::ValueNumber;
+
+
+
+pub type TesterFn = fn(
```
SergioBenitez: This isn't idiomatic Rust style, and in this case, the idiomatic Rust style takes up just as many lines, so it's better to use that.

src/testers.rs
```diff
@@ -1,30 +1,123 @@
...
+    try!(number_args_allowed("undefined", 0, params.len()));
+
+    Ok(value.is_none())
+}
+
+/// Returns true if `value` is a string. Otherwise, returns false.
+pub fn string(value: Option<Value>, params: Vec<Value>) -> TeraResult<bool> {
+    try!(number_args_allowed("string", 0, params.len()));
+
+    match value {
```
SergioBenitez: This can be better written as:

```rust
match value {
    Some(Value::String(_)) => true,
    None => value_is_undefined("string"),
    _ => false
}
```

src/testers.rs
```diff
@@ -1,30 +1,123 @@
...
-    let name = match *value {
-        Node::Identifier { ref name, .. } => name,
-        _ => return Err(TeraError::TestError("defined".to_string(),
-                "defined can only be called on identifiers".to_string()))
-    };
+    Ok(())
+}
+
+// Called to return an error when unwrapping the value and realising there's nothing
+fn value_is_undefined(name: &str) -> TeraResult<bool> {
```
SergioBenitez: This should take the name of the variable tested as well.

src/testers.rs
```diff
@@ -1,30 +1,123 @@
...
+    Err(TeraError::TestError(
+        name.to_string(),
+        format!("{} was called on an undefined expression", name)
```
SergioBenitez: And the name should be in the error message: "`{}` was called on the variable `{}`, which is undefined."

src/testers.rs
```diff
@@ -1,30 +1,123 @@
...
+        },
+        None => value_is_undefined("string")
+    }
+}
+
+/// Returns true if `value` is a number. Otherwise, returns false.
+pub fn number(value: Option<Value>, params: Vec<Value>) -> TeraResult<bool> {
+    try!(number_args_allowed("number", 0, params.len()));
+
+    match value {
```
SergioBenitez: Same refactoring here.

SergioBenitez:
Looks pretty good overall! Left some comments here and there.

src/render.rs
Keats: The `try!` call would return in the closure and not the parent function no ?

src/testers.rs
Keats: what would be the idiomatic style? I'm just using the style I use in python

src/testers.rs
Keats: Yep, need to add the name of the variable to the tests fn definition

src/render.rs
SergioBenitez: Ah indeed. Disregard this.

src/testers.rs
Keats: I find that a bit harder to read myself but it is neater

src/testers.rs
SergioBenitez: When the line length exceeds some number of characters, the params should line up:

```rust
pub type TesterFn = fn(value: Option<Value>,
                       params: Vec<Value>)
                       -> TeraResult<bool>;
```

But you might be able to fit the following here:

```rust
pub type TesterFn = fn(Option<Value>, Vec<Value>) -> TeraResult<bool>;
```

Keats:
Updated the PR, thanks for the review @SergioBenitez !

src/testers.rs
```diff
@@ -1,30 +1,130 @@
...
+        _ => Ok(false)
+    }
 }
 
+/// Returns true if `value` is an odd number. Otherwise, returns false.
+pub fn odd(name: &str, value: Option<Value>, params: Vec<Value>) -> TeraResult<bool> {
+    try!(number_args_allowed(name, "odd", 0, params.len()));
+
+    match value {
+        Some(v) => match v.to_number() {
```
SergioBenitez: I think the `value_is_undefined` call should be renamed `value_defined` and used as in a try expression. Then, this function (and others) can be rewritten like:

```rust
try!(number_args_allowed(name, "odd", 0, params.len()));
try!(value_defined(name, "odd"));

match value.and_then(|v| v.to_number()) {
    Some(number) => Ok(number % 2.0 != 0.0),
    None => return Err(TeraError::TestError(
        "odd".into(), "odd can only be called on numbers".into())
    ),
}
```

I think this looks significantly cleaner.

src/testers.rs
```diff
@@ -1,30 +1,130 @@
...
+                "odd can only be called on numbers".to_string()
+            ))
+        },
+        None => value_is_undefined(name, "odd")
+    }
+}
+
+
+/// Returns true if `value` is an even number. Otherwise, returns false.
+pub fn even(name: &str, value: Option<Value>, params: Vec<Value>) -> TeraResult<bool> {
```
SergioBenitez: This should just be:

```rust
odd(name, value, params).map(|b| !b)
```

or, for clarity:

```rust
let is_odd = try!(odd(name, value, params);
!is_odd
```

I like the one-liner myself.

src/testers.rs
```diff
@@ -58,63 +62,52 @@ pub fn undefined(name: &str, value: Option<Value>, params: Vec<Value>) -> TeraRe
 /// Returns true if `value` is a string. Otherwise, returns false.
 pub fn string(name: &str, value: Option<Value>, params: Vec<Value>) -> TeraResult<bool> {
     try!(number_args_allowed(name, "string", 0, params.len()));
+    try!(value_defined(name, "string", &value));
 
-    match value {
-        Some(Value::String(_)) => Ok(true),
-        None => value_is_undefined(name, "string"),
+    match value.unwrap() {
```
SergioBenitez: You should try to never call `unwrap`. This is considered to be an anti-pattern in Rust. What if the `try!` call above gets removed in some later version, for example? At the very least, use `expect`.

src/testers.rs
```diff
@@ -58,63 +62,52 @@ pub fn undefined(name: &str, value: Option<Value>, params: Vec<Value>) -> TeraRe
...
 
 /// Returns true if `value` is a number. Otherwise, returns false.
 pub fn number(name: &str, value: Option<Value>, params: Vec<Value>) -> TeraResult<bool> {
     try!(number_args_allowed(name, "number", 0, params.len()));
+    try!(value_defined(name, "number", &value));
 
-    match value {
-        Some(Value::I64(_)) | Some(Value::F64(_)) | Some(Value::U64(_)) => Ok(true),
-        None => value_is_undefined(name, "number"),
+    match value.unwrap() {
```
SergioBenitez: Same.

src/testers.rs
```diff
@@ -58,63 +62,52 @@ pub fn undefined(name: &str, value: Option<Value>, params: Vec<Value>) -> TeraRe
...
-    match value {
-        Some(v) => match v.to_number() {
-            Ok(f) => Ok(f % 2.0 != 0.0),
-            Err(_) => Err(TeraError::TestError(
-                "odd".to_string(),
-                "odd can only be called on numbers".to_string()
-            ))
-        },
-        None => value_is_undefined(name, "odd")
+    match value.unwrap().to_number() {
```
SergioBenitez: Same, and for the rest.

src/testers.rs
Keats: I disagree on that, if the unwrap is safe (like here) I don't mind it. That or `expect`. Test(s) cover the case where the line above is removed.

I'll change it though, the `and_then` approach is neater

Issue #45, opened 2016-08-25
Use our own serialization


Keats:
rather than JSON.
It would give us more types than JSON has, like date and datetimes out of the box.

Would require custom derive to be nice though so waiting for macros 1.1 to land for that to happen.
Or we could piggyback on Serde and write a `serde_tera` implementation

Alex-PK:
I've been working on something like that for a while now, but mostly to experiment with lower-level constructs in Rust. It's very much "in flux" as I learn new things and conduct new experiments:

https://bitbucket.org/Alex_PK/darry/overview

The primary policy is to make it always work on stable without requiring nightly features, and to use as few macros as possible, so I am a bit limited in what I can do, but I kind of like what's coming out. :)

Keats:
Interesting! i think I'd go with a fork of `serde_json` though and add whatever type we want though, as it will be the way to serialize so we should get all the speed with no efforts.

I had a quick look at the code and I'd rename https://bitbucket.org/Alex_PK/darry/src/ad0cc393549be8e558e349790e968ac76ffe2b74/src/value.rs?at=master&fileviewer=file-view-default#value.rs-137 to `is_truthy` or something similar, i'd expect a `to_bool` method to fail on something other than a bool

Keats:
`toml-rs` (http://alexcrichton.com/toml-rs/toml/index.html) switched to something like that.
Would be interesting to try it out on a branch

PR #65, opened 2016-10-09
Expose renderer directly to allow non-Context JSON object values.

SergioBenitez:
It would be nice to render JSON values directly without first converting into a `Context` object. This is the way other Rust templating libraries work, and I think it makes the most sense. One way to do this is to export the Renderer type so that someone can call `Renderer::new_with_json` and then `render` directly. This is what this PR contributes.

### DIFF
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -4,6 +4,8 @@ use std::fs::File;
 use std::fmt;
 
 use glob::glob;
+use serde::Serialize;
+use serde_json::value::to_value;
 
 use template::Template;
 use filters::{FilterFn, string, array, common};
@@ -59,8 +61,22 @@ impl Tera {
 
     pub fn render(&self, template_name: &str, data: Context) -> TeraResult<String> {
         let template = try!(self.get_template(template_name));
-        let mut renderer = Renderer::new(template, self, data);
+        let mut renderer = Renderer::new(template, self, data.as_json());
 
+        renderer.render()
+    }
+
+    /// Renders a Serde JSON value.
+    pub fn render_value<T>(&self, template_name: &str, data: &T) -> TeraResult<String>
+        where T: Serialize
+    {
+        let value = to_value(data);
+        if !value.is_object() {
+            return Err(TeraError::InvalidValue(template_name.to_string()))
+        }
+
+        let template = try!(self.get_template(template_name));
+        let mut renderer = Renderer::new(template, self, value);
         renderer.render()
     }
 
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -59,6 +59,7 @@ impl Tera {
         tera
     }
 
+    /// Renders a Tera template given a `Context`.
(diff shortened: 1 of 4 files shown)


Keats:
I'm not convinced about the `Renderer::new_with_json` as that allows invalid calls, `render_with_json(..,..,to_value(&1)` would be valid. An alternative would be a `render` method that takes `BTreeMap<String, Value>` as arg but i'm not sure it's valuable at all.

Another thing against the `render_with_json` is that json is a temporary format for Tera, hopefully we will be able to move to something else (#45)

Also if it exists it should be on the `Tera` instance like the `render`

SergioBenitez:
Restricting rendering to `Context` (or any non-polymorphic type, to be clear) is a considerable usability problem. Imagine a library that wants to expose an interface to Tera. With the current design, and the one you've proposed, the library is shoehorned into making one of two choices: 1) it must copy everything into a `Context`, or 2) it has to re-export `Context` and force their users into using that type.

What's the issue with allowing something that's not a `Value::Object` to be passed in? If that happens, we should simply return an error to the user. That is, `new_with_json ` should return a `Result` object if the Value isn't an object. I didn't do that here because later code (currently, though you've removed it in a PR) checks if the value is indeed a `Value::Object`.

Alternatively, you can define your own serializer and deserializer for Serde and ship that along with Tera. But this is functionally identical to what I've proposed, except it moves the `Result` up a level to the serialization step.

Keats:
The last paragraph is the optimal path as it allows more information than what JSON offers (dates etc) but it is not an immediate goal.

You can change the type of the `Renderer` constructor to take `context` as a value and call `to_json` in `Tera::render`. This way `Tera::render_with_json` is easier as it only needs to check if the `Value` is of type `Value::Object` and then only calls the renderer with correct input.

SergioBenitez:
Sounds good to me. Added `Tera::render_value`, which I thought sounded a bit better than `Tera::render_with_json`.

SergioBenitez:
Actually, that sounds like it's rendering the value itself, so I changed it to `value_render`.

src/tera.rs
```diff
@@ -57,13 +59,28 @@ impl Tera {
...
     pub fn render(&self, template_name: &str, data: Context) -> TeraResult<String> {
         let template = try!(self.get_template(template_name));
-        let mut renderer = Renderer::new(template, self, data);
+        let mut renderer = Renderer::new(template, self, data.as_json());
 
         renderer.render()
     }
 
+    /// Renders a Tera template given a `Serializeable` object.
+    pub fn value_render<T>(&self, template_name: &str, data: &T) -> TeraResult<String>
```
Keats: `render_from_value`?

Keats:
Only missing integrations tests for render failures and render ok  for the new `render` method + some docs on README and it's ready to go

Keats:
I'll do it, thanks!

Issue #66, opened 2016-10-09
Macros


Keats:
Macros are a way to do what most templates engine do as include but with a known context as they look like functions.

See [Jinja2 docs](http://jinja.pocoo.org/docs/dev/templates/#macros) and some [real life examples](https://github.com/pallets/website/tree/master/templates/macros).

A few things to consider when porting them:
- import
- `call` syntax
- args
## Import

Jinja2 has 2 ways to load macros (http://jinja.pocoo.org/docs/dev/templates/#import):

```
{% import "macros.html" as macros %}
{% from "macros.html" import macro_1 , macro_2 %}
```

Since the equivalent of `from x import *` is not great in general, I'd be interested in the second form, that is `{% from "macros.html" import macro_1 , macro_2 %}` which makes it clear what you want to import. 
That means `Template` need to have a `pub macros: HashMap<String, Node>` or something similar in order to resolve the macros easily. 
Since we are namespacing macros by filename, several macros can have the same name which can cause an issue if we are trying to use 2 macros with the same name in the same file.
Jinja2 allows renaming imports `{% from 'forms.html' import input as input_field, textarea %}
` and we could/should support that syntax, waiting for feedback on that point
## Call syntax

I'm not sure `call` is that needed and it makes things a bit hard to understand imo. For example from the Jinja2 docs:

```
{% macro dump_users(users) -%}
    <ul>
    {%- for user in users %}
        <li><p>{{ user.username|e }}</p>{{ caller(user) }}</li>
    {%- endfor %}
    </ul>
{%- endmacro %}

{% call(user) dump_users(list_of_user) %}
    <dl>
        <dl>Realname</dl>
        <dd>{{ user.realname|e }}</dd>
        <dl>Description</dl>
        <dd>{{ user.description }}</dd>
    </dl>
{% endcall %}
```

seems overly complicated to read for what should be simple.
Let's ignore `call` for the first pass of macros.
## Args

To match what the filters are doing in Tera, all args to a macros should be kwargs.
## Motivation for macros

From the creator of Jinja2:

> Another thing i strongly disagree with but I can see the motivation. Macros in Jinja I prefer so much over includes because it becomes clear what values exist. Includes become really messy unless you are very organized. Macros (which are just functions) accept the parameters and it becomes clear what the thing does. Just look at that macro as an example: https://github.com/pallets/website/blob/master/templates/mac...
[...]

PR #69, opened 2016-10-19
Macros

Keats:
Also uniformize function call args between macros and filters

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,12 +378,12 @@ Macros are defined as follows:
 {% endmacro hello_world %}
 ```
 
-In order to be called, the file containing the macros needs to be loaded first in the file using them:
+You need to import the file containing the macros in order to use them:
 
 ```jinja2
 {% import "macros.html" as macros %}
 ```
-You can name your namespace (`macros` in the example) anything you want.
+You can name that file namespace (`macros` in the example) anything you want.
 You can call a macro the following way:
 
 ```jinja2
@@ -391,3 +391,7 @@ You can call a macro the following way:
 ```
 Do note that macros, like other functions in Tera, require keyword arguments.
 If you are trying to call a macro defined in the same file, you will need to use the `{{ self::my_macro() }}`` syntax to call it the file its defined in.
+
+Macros can be called recursively but there is no limit to recursion so make sure you macro ends.
+
+There are a few restrictions to the content inside macros: no macros definitions, includes or blocks.
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ For example, here's a `base.html` almost copied from the jinja documentation:
 </html>
 ```
 The difference with Jinja being that `endblock` tags must be named.
-This defines 4 `block` tag that child templates can override. The `head` and `footer` block contains some html already which will be rendered if they are not overrident.
+This defines 4 `block` tag that child templates can override. The `head` and `footer` block contains some html already which will be rendered if they are not overriden.
 
 #### Child template
 Again, straight from jinja2 docs:
--- a/README.md
+++ b/README.md
@@ -394,4 +394,4 @@ If you are trying to call a macro defined in the same file, you will need to use
 
 Macros can be called recursively but there is no limit to recursion so make sure you macro ends.
 
-There are a few restrictions to the content inside macros: no macros definitions, includes or blocks.
+There are a few restrictions to the content inside macros: no macros definitions or blocks are allowed.
--- a/README.md
+++ b/README.md
@@ -365,16 +365,16 @@ Escapes a string's HTML. Specifically, it makes these replacements:
 - & is converted to &amp;
 
 ### Macros
-Macros are a simple way to re-use template bits. Think of them as functions that you can call
-and that returns some text.
+Macros are a simple way to reuse template bits. Think of them as functions that you can call that return some text.
 
-Macros are defined the following way:
+Macros are defined as follows:
 
 ```jinja2
-{% macro hello_world(greeting, language) %}
-    {% if language == "en" %}
-        {{ greeting }}
-    {% endif %}
+{% macro input(label, type) %}
+    <label>
+        {{ label }}
+        <input type="{{type}}" />
+    </label>
 {% endmacro hello_world %}
 ```
 
@@ -387,7 +387,7 @@ You can name your namespace (`macros` in the example) anything you want.
 You can call a macro the following way:
 
 ```jinja2
-{{ macros::hello_world(greeting="Hello World", language="en") }}
+{{ macros::input(label="Name", type="text") }}
 ```
 Do note that macros, like other functions in Tera, require keyword arguments.
 If you are trying to call a macro defined in the same file, you will need to use the `{{ self::my_macro() }}`` syntax to call it the file its defined in.
--- a/README.md
+++ b/README.md
@@ -364,3 +364,30 @@ Escapes a string's HTML. Specifically, it makes these replacements:
 - " (double quote) is converted to &quot;
 - & is converted to &amp;
 
+### Macros
+Macros are a simple way to re-use template bits. Think of them as functions that you can call
+and that returns some text.
+
+Macros are defined the following way:
+
+```jinja2
+{% macro hello_world(greeting, language) %}
+    {% if language == "en" %}
+        {{ greeting }}
+    {% endif %}
+{% endmacro hello_world %}
+```
+
+In order to be called, the file containing the macros needs to be loaded first in the file using them:
+
+```jinja2
+{% import "macros.html" as macros %}
+```
+You can name your namespace (`macros` in the example) anything you want.
+You can call a macro the following way:
+
+```jinja2
+{{ macros::hello_world(greeting="Hello World", language="en") }}
+```
+Do note that macros, like other functions in Tera, require keyword arguments.
+If you are trying to call a macro defined in the same file, you will need to use the `{{ self::my_macro() }}`` syntax to call it the file its defined in.
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,3 +1,5 @@
+use std::collections::{LinkedList};
+
 use serde_json::value::Value;
 
 /// Library generic result type.
@@ -69,7 +71,7 @@ quick_error! {
             display("Tester `{}` encountered an error while running: {}", tester_name, message)
             description("tester runtime error")
         }
-        MacroCallWrongArgs(macro_name: String, expected_args: Vec<String>, args: Vec<String>) {
+        MacroCallWrongArgs(macro_name: String, expected_args: LinkedList<String>, args: Vec<String>) {
             display("Macro `{}` got `{:?}` for args but was expecting `{:?}` (order does not matter)", macro_name, expected_args, args)
             description("macro wrong args")
         }
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -33,8 +33,8 @@ quick_error! {
             display("Filter `{}` was not found.", name)
             description("filter not found")
         }
-        MacroNotFound(name: String, tpl_name: String) {
-            display("Macro `{}` was not found in the namespace `{}`.", name, tpl_name)
+        MacroNotFound(name: String, namespace: String) {
+            display("Macro `{}` was not found in the namespace `{}`.", name, namespace)
             description("macro not found")
         }
         NotANumber(name: String) {
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -9,7 +9,7 @@ quick_error! {
         MismatchingEndTag(line_no: usize, col_no: usize, expected: String, found: String) {
             display("Was expecting block with name `{}` to be closed, but `{}` is closing at line {:?}, column {:?}",
                     expected, found, line_no, col_no)
-            description("unexpected endX name")
+            description("unexpected end tag name")
         }
         InvalidSyntax(line_no: usize, col_no: usize) {
             display("invalid Tera syntax at line {:?}, column {:?}", line_no, col_no)
@@ -26,7 +26,7 @@ quick_error! {
             description("invalid value")
         }
         TemplateNotFound(name: String) {
-            display("Template `{}` wasn't found", name)
+            display("Template `{}` was not found", name)
             description("template not found")
         }
         FilterNotFound(name: String) {
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -6,10 +6,10 @@ pub type TeraResult<T> = Result<T, TeraError>;
 quick_error! {
     #[derive(PartialEq, Debug, Clone)]
     pub enum TeraError {
-        MismatchingEndBlock(line_no: usize, col_no: usize, expected: String, found: String) {
-            display("Was expecting block `{}` to be closed, but `{}` is closing at line {:?}, column {:?}",
+        MismatchingEndTag(line_no: usize, col_no: usize, expected: String, found: String) {
+            display("Was expecting block with name `{}` to be closed, but `{}` is closing at line {:?}, column {:?}",
                     expected, found, line_no, col_no)
-            description("unexpected endblock name")
+            description("unexpected endX name")
         }
         InvalidSyntax(line_no: usize, col_no: usize) {
             display("invalid Tera syntax at line {:?}, column {:?}", line_no, col_no)
@@ -19,20 +19,24 @@ quick_error! {
             display("deprecated syntax at line {:?}, column {:?}: {}", line_no, col_no, message)
             description("deprecated syntax")
         }
-        TemplateNotFound(name: String) {
-            display("Template `{}` wasn't found", name)
-            description("template not found")
-        }
+
+        // Runtime errors
         InvalidValue(name: String) {
             display("Expected the value to be an Object while rendering `{}`.", name)
             description("invalid value")
         }
+        TemplateNotFound(name: String) {
+            display("Template `{}` wasn't found", name)
+            description("template not found")
+        }
         FilterNotFound(name: String) {
-            display("Filter `{}` was not found in the context.", name)
+            display("Filter `{}` was not found.", name)
             description("filter not found")
         }
-
-        // Runtime errors
+        MacroNotFound(name: String, tpl_name: String) {
+            display("Macro `{}` was not found in the namespace `{}`.", name, tpl_name)
+            description("macro not found")
+        }
         NotANumber(name: String) {
             display("Field `{}` was used in a math operation but is not a number", name)
             description("field is not a number")
@@ -57,16 +61,17 @@ quick_error! {
             display("Filter `{}` expected an arg called `{}`", filter_name, arg_name)
             description("missing arg in filter call")
         }
-
-        // Tester errors.
         TesterNotFound(name: String) {
             display("Tester `{}` was not found in the context.", name)
             description("tester not found")
         }
         TestError(tester_name: String, message: String) {
-            display("Tester `{}` encountered an error while running: {}",
-                    tester_name, message)
+            display("Tester `{}` encountered an error while running: {}", tester_name, message)
             description("tester runtime error")
+        }
+        MacroCallWrongArgs(macro_name: String, expected_args: Vec<String>, args: Vec<String>) {
+            display("Macro `{}` got `{:?}` for args but was expecting `{:?}` (order does not matter)", macro_name, expected_args, args)
+            description("macro wrong args")
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,7 +1,7 @@
 #![allow(dead_code)]
 
 // Needed by pest
-#![recursion_limit = "200"]
+#![recursion_limit = "300"]
 
 #![cfg_attr(feature = "dev", feature(plugin))]
 #![cfg_attr(feature = "dev", plugin(clippy))]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,8 +31,7 @@ mod testers;
 // Library exports.
 
 // Template is meant to be used internally only but is exported for test/bench.
-#[doc(hidden)]
-pub use template::Template;
+#[doc(hidden)] pub use template::Template;
 pub use context::Context;
 pub use tera::Tera;
 pub use errors::{TeraResult, TeraError};
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -10,19 +10,15 @@
 extern crate serde;
 extern crate serde_json;
 extern crate glob;
-#[macro_use]
-extern crate pest;
-#[macro_use]
-extern crate quick_error;
+#[macro_use] extern crate pest;
+#[macro_use] extern crate quick_error;
 extern crate slug;
 extern crate regex;
-#[macro_use]
-extern crate lazy_static;
+#[macro_use] extern crate lazy_static;
 extern crate url;
 
 mod errors;
-#[macro_use]
-mod macros;
+#[macro_use] mod macros;
 mod parser;
 mod context;
 mod render;
@@ -35,7 +31,8 @@ mod testers;
 // Library exports.
 
 // Template is meant to be used internally only but is exported for test/bench.
-#[doc(hidden)] pub use template::Template;
+#[doc(hidden)]
+pub use template::Template;
 pub use context::Context;
 pub use tera::Tera;
 pub use errors::{TeraResult, TeraError};
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -114,7 +114,7 @@ impl_rdp! {
         // TODO: add default arg?
         macro_param = @{ simple_ident }
         macro_params = !@{ macro_param ~ ([","] ~ macro_param )* }
-        macro_definition = _{ identifier ~ ["("] ~ macro_params* ~ [")"]}
+        macro_definition = _{ identifier ~ ["("] ~ macro_params? ~ [")"]}
         macro_call = { simple_ident ~ ["::"] ~ simple_ident ~ ["("] ~ fn_args? ~ [")"] }
 
         // Variable tests.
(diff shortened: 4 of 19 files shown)


Keats:
@SergioBenitez @andrelmartins 
Roughly working macro system based on discusssion in #66 

A few interesting things popped up when working on it:

- current system does not allow to use macro in the same file they are defined, we could use something like `{{ self::my_macro() }}` where `self` resolves to the current template file
- not sure whether to have default args for macros
- going to uniformize args in general so they will all be kwargs, including tests which mean I will change the current syntax and remove the `is divisibleby 3` syntax
- i don't think it's going to be possible to have `include` use macro
- the `namespace::` syntax breaks the Jinja2 highlighting but I think it's worth it

andrelmartins:
@Keats Sorry for delay in replying, wanted to get a chance to test it out.

The current implementation seems nice enough to use, at least for me.

I'm in favor of adding the "self::macro()" variant, since that opens the door to more complex macro code and to adding one-off macros to simplify repeated code that's specific to a given page.

As for default arguments, given that you moved macros to kwargs, it seems like a useful next step to implement default arguments. But it is not a significant feature in my view, just a nice to have.

andrelmartins:
Sorry for delay in replying, wanted to get a chance to test it out.

The current implementation seems nice enough to use, at least for me.

I'm in favor of adding the "self::macro()" variant, since that opens the
door to more complex macro code and to adding one-off macros to simplify
repeated code that's specific to a given page.

As for default arguments, given that you moved macros to kwargs, it seems
like a useful next step to implement default arguments. But it is not a
significant feature in my view, just a nice to have.



On Wed, Oct 19, 2016 at 3:37 AM, Vincent Prouillet <notifications@github.com

Keats:
@andrelmartins no worries i'm still in the middle of moving/vacation so I'll finish it up next week.

I guess those macros don't replace your need for `include` though?

andrelmartins:
@Keats Correct, I still need the `include` tag as is.

Keats:
Added the `self::` syntax but the whole thing needs cleaning and testing before being mergeable

SergioBenitez:
@Keats Just had a chance to look at this, albeit only in a cursory way. Looks fantastic! Great job! Happy to give it a more thorough look once you think it's ready for merging.

Keats:
Looking at some tests examples, kwargs would be annoying to use with them and `is oneof(a, b, c)` wouldn't really work so probably not putting kwargs for tests

Keats:
@SergioBenitez feel free to do a full review when you have time, I don't think I'll change it significantly

README.md
```diff
@@ -364,3 +364,30 @@ Escapes a string's HTML. Specifically, it makes these replacements:
 - " (double quote) is converted to &quot;
 - & is converted to &amp;
 
+### Macros
+Macros are a simple way to re-use template bits. Think of them as functions that you can call
```
SergioBenitez: It's "reuse", not "re-use".

README.md
```diff
@@ -364,3 +364,30 @@ Escapes a string's HTML. Specifically, it makes these replacements:
...
+and that returns some text.
```
SergioBenitez: "and" shouldn't be there -> "that return text".

README.md
```diff
@@ -364,3 +364,30 @@ Escapes a string's HTML. Specifically, it makes these replacements:
...
+
+Macros are defined the following way:
```
SergioBenitez: "the following way" -> "as follows"

README.md
```diff
@@ -364,3 +364,30 @@ Escapes a string's HTML. Specifically, it makes these replacements:
...
+
+```jinja2
+{% macro hello_world(greeting, language) %}
```
SergioBenitez: This is a bit of a weird example since it won't emit anything if language is not equal to "en". Maybe you can add some "else if"s or maybe just an "else"?

README.md
```diff
@@ -364,3 +364,30 @@ Escapes a string's HTML. Specifically, it makes these replacements:
...
+    {% if language == "en" %}
+        {{ greeting }}
+    {% endif %}
+{% endmacro hello_world %}
+```
+
+In order to be called, the file containing the macros needs to be loaded first in the file using them:
```
SergioBenitez: The wording here is a bit odd.

README.md
```diff
@@ -364,3 +364,30 @@ Escapes a string's HTML. Specifically, it makes these replacements:
...
+```jinja2
+{% import "macros.html" as macros %}
+```
+You can name your namespace (`macros` in the example) anything you want.
+You can call a macro the following way:
+
+```jinja2
+{{ macros::hello_world(greeting="Hello World", language="en") }}
+```
+Do note that macros, like other functions in Tera, require keyword arguments.
```
SergioBenitez: Why is this the case? Positional arguments are nice in some cases.

src/errors.rs
```diff
@@ -6,10 +6,10 @@ pub type TeraResult<T> = Result<T, TeraError>;
 quick_error! {
     #[derive(PartialEq, Debug, Clone)]
     pub enum TeraError {
-        MismatchingEndBlock(line_no: usize, col_no: usize, expected: String, found: String) {
-            display("Was expecting block `{}` to be closed, but `{}` is closing at line {:?}, column {:?}",
+        MismatchingEndTag(line_no: usize, col_no: usize, expected: String, found: String) {
+            display("Was expecting block with name `{}` to be closed, but `{}` is closing at line {:?}, column {:?}",
                     expected, found, line_no, col_no)
-            description("unexpected endblock name")
+            description("unexpected endX name")
```
SergioBenitez: What is "endX"?

src/errors.rs
```diff
@@ -19,20 +19,24 @@ quick_error! {
...
-            description("template not found")
-        }
+
+        // Runtime errors
         InvalidValue(name: String) {
             display("Expected the value to be an Object while rendering `{}`.", name)
             description("invalid value")
         }
+        TemplateNotFound(name: String) {
+            display("Template `{}` wasn't found", name)
```
SergioBenitez: "wasn't" -> "was not"

src/lib.rs
```diff
@@ -35,7 +31,8 @@ mod testers;
 // Library exports.
 
 // Template is meant to be used internally only but is exported for test/bench.
-#[doc(hidden)] pub use template::Template;
+#[doc(hidden)]
```
SergioBenitez: Why not inline this attribute as well? The others were inlined.

src/parser.rs
```diff
@@ -104,6 +110,13 @@ impl_rdp! {
         identifier_with_filter = { identifier ~ filters }
         idents = _{ identifier_with_filter | identifier }
 
+        // macros
+        // TODO: add default arg?
+        macro_param = @{ simple_ident }
+        macro_params = !@{ macro_param ~ ([","] ~ macro_param )* }
+        macro_definition = _{ identifier ~ ["("] ~ macro_params* ~ [")"]}
```
SergioBenitez: Shouldn't the `*` be a `?`? `macro_params` already does the repetition.

src/parser.rs
```diff
@@ -136,29 +149,34 @@ impl_rdp! {
...
-        elif_tag        = !@{ tag_start ~ ["elif"] ~ expression ~ test? ~ tag_end }
-        else_tag        = !@{ tag_start ~ ["else"] ~ tag_end }
-        for_tag         = !@{ tag_start ~ ["for"] ~ identifier ~ ["in"] ~ idents ~ tag_end }
-        raw_tag         = !@{ tag_start ~ ["raw"] ~ tag_end }
-        endraw_tag      = !@{ tag_start ~ ["endraw"] ~ tag_end }
-        endblock_tag    = !@{ tag_start ~ ["endblock"] ~ identifier ~ tag_end }
-        endif_tag       = !@{ tag_start ~ ["endif"] ~ tag_end }
-        endfor_tag      = !@{ tag_start ~ ["endfor"] ~ tag_end }
+        include_tag      = !@{ tag_start ~ ["include"] ~ string ~ tag_end }
+        import_macro_tag = !@{ tag_start ~ ["import"] ~ string ~ ["as"] ~ simple_ident ~ tag_end}
```
SergioBenitez: Why is the "as ..." part always required?

src/parser.rs
```diff
@@ -210,6 +228,19 @@ impl_rdp! {
             (_: include_tag, &name: string) => {
                 Ok(Some(Node::Include(name.replace("\"", "").to_string())))
             },
+            (_: import_macro_tag, &tpl_name: string, &name: simple_ident) => {
+                Ok(Some(Node::ImportMacro {
+                    tpl_name: tpl_name.replace("\"", "").to_string(),
```
SergioBenitez: `tpl_name.trim_matches("\"").to_string()` conveys more information, I think. It's also faster.

src/parser.rs
```diff
@@ -912,13 +985,13 @@ mod tests {
 
     #[test]
     fn test_ast_if_with_test_params() {
-        let parsed_ast = parse(r#"{% if pi is equalto 3.14 %}Hey{% endif %}"#);
+        let parsed_ast = parse(r#"{% if pi is equalto 3.13 %}Hey{% endif %}"#);
```
SergioBenitez: Why this change?

src/parser.rs
```diff
@@ -1062,6 +1135,7 @@ mod tests {
 
     #[test]
     fn test_unary_test() {
+        // TODO: remove that syntax
```
SergioBenitez: Why? It's very convenient.

src/parser.rs
```diff
@@ -1078,6 +1152,7 @@ mod tests {
         assert!(parser.end());
     }
 
+    // TODO: remove that syntax
```
SergioBenitez: This test checks that the syntax there _doesn't_ work.

src/parser.rs
```diff
@@ -1132,6 +1205,60 @@ mod tests {
     }
 
     #[test]
+    fn test_ast_macro_definition() {
+        let parsed_ast = parse("{% macro helloworld(greeting) %}Hello{% endmacro helloworld %}");
```
SergioBenitez: There should probably be tests that use zero parameters and more than one parameter.

src/parser.rs
```diff
@@ -1132,6 +1205,60 @@ mod tests {
...
+            name: "macros".to_string(),
+        });
+
+        let root = Node::List(ast);
+        assert_eq!(parsed_ast.unwrap(), root);
+    }
+
+    #[test]
+    fn test_ast_macro_call() {
+        let parsed_ast = parse("{{ macros::macro1(foo=bar, hey=1+2) }}");
```
SergioBenitez: Likewise about the arity on the calling side.

src/render.rs
```diff
@@ -345,6 +363,71 @@ impl<'a> Renderer<'a> {
         Ok(output.trim_right().to_string())
     }
 
+    // TODO: have a look at this method once it is working to clean it up, too much indentation
+    // Probably very inefficient and can be trimmed down
+    fn render_macro(&mut self, call_node: Node) -> TeraResult<String> {
+        match call_node {
```
SergioBenitez: Use `if let` instead.

src/render.rs
```diff
@@ -345,6 +363,71 @@ impl<'a> Renderer<'a> {
...
+            MacroCall {namespace, name, params} => {
+                // Avoid shadowing when matching the definition of macros
+                let call_params = params;
```
SergioBenitez: Not sure why this is done.

src/render.rs
```diff
@@ -345,6 +363,71 @@ impl<'a> Renderer<'a> {
...
+                // To find the real namespace we just look at the last one we got.
+                // If we have self, we take the last one otherwise we push it into the vec
+                let real_namespace = if namespace == "self" {
+                    // TODO: handle error if we don't have a namespace
+                    self.macro_namespaces
+                        .last()
+                        .expect("Open an issue with a template sample please (mention `self namespace macro`)!")
+                        .to_string()
+                } else {
+                    self.macro_namespaces.push(namespace.clone());
```
SergioBenitez: Do you really need to clone twice?

src/render.rs
```diff
@@ -345,6 +363,71 @@ impl<'a> Renderer<'a> {
...
+    // TODO: have a look at this method once it is working to clean it up, too much indentation
+    // Probably very inefficient and can be trimmed down
+    fn render_macro(&mut self, call_node: Node) -> TeraResult<String> {
+        match call_node {
+            MacroCall {namespace, name, params} => {
+                // Avoid shadowing when matching the definition of macros
+                let call_params = params;
+                // To find the real namespace we just look at the last one we got.
+                // If we have self, we take the last one otherwise we push it into the vec
+                let real_namespace = if namespace == "self" {
```
SergioBenitez: A `match` would be cleaner here.

src/render.rs
```diff
@@ -345,6 +363,71 @@ impl<'a> Renderer<'a> {
...
+                    self.macro_namespaces
+                        .last()
+                        .expect("Open an issue with a template sample please (mention `self namespace macro`)!")
+                        .to_string()
+                } else {
+                    self.macro_namespaces.push(namespace.clone());
+                    namespace.clone()
+                };
+                // TODO: avoid cloned() everywhere if possible
+                match self.macros.get(&real_namespace).cloned() {
```
SergioBenitez: This is very, very hairy, and likely needlessly so. Use the `Option` combinators to make it much more accessible: `map`, `and_then`, etc.

src/render.rs
```diff
@@ -356,6 +439,14 @@ impl<'a> Renderer<'a> {
 
                 Ok(output)
             },
+            ImportMacro {tpl_name, name} => {
+                let tpl = try!(self.tera.get_template(&tpl_name));
+                self.macros.insert(name.to_string(), tpl.macros.clone());
+                // In theory, the render_node should return TeraResult<Option<String>>
+                // but in practice there's no difference so keeping this hack
```
SergioBenitez: Except this allocates needlessly. Just do `let x = match node` at the top. Here, `return Ok(None)`. Everywhere else, remove the `Ok(..)`, but keep what's inside. Then add a final `Ok(x)` to the end of the function.

src/template.rs
```diff
@@ -81,4 +107,12 @@ mod tests {
         assert_eq!(tpl.parent.unwrap(), "base.html".to_string());
         assert_eq!(tpl.blocks.contains_key("hey"), true);
     }
+
+    #[test]
+    fn test_can_find_macros() {
```
SergioBenitez: This is a fairly weak test.

tests/expected/use_macros.html
```diff
@@ -0,0 +1,5 @@
+
```
SergioBenitez: Why is there so much empty space? Probably don't want this.

tests/render-failures/macros.html
```diff
@@ -0,0 +1,10 @@
+{% macro hello_world(greeting, language) %}
+    {% if language == "en" %}
```
SergioBenitez: Same as my comment in the README: this is strange example.

README.md
Keats: Experiment. I want to see how that works in practice and I'll change if it feels annoying

src/lib.rs
Keats: Intellij rust auto format, i'll open an issue on it in a bit for that

src/parser.rs
Keats: to avoid clippy warning about pi

src/parser.rs
Keats: To have a namespace for the macros fns, see the post by @andrelmartins on https://github.com/Keats/tera/issues/66

Keats:
@SergioBenitez thanks for the comments!
I'm still not sure what to allow with `self`, I had trouble explaning it in the docs which is not really a good sign. I'll try to come up with a better design this week. Another solution would be to ship an initial version macros that doesn't allow macro calls at the beginning

src/render.rs
Keats: I didn't realise I could use `if let` for enum variant!

src/template.rs
Keats: granted it doesn't test that much but it's still helpful I think

Keats:
Updated the code, thanks for the comments @SergioBenitez 
Still need to think through a bit more the `self` and similar things

Keats:
> I feel that some interesting test cases are missing. What happens if a macro is recursive? Should Tera detect that or not? How many levels deep can I nest things? Can I pass macro invocations to macros? Can I pass macro names to macros? A lot of questions are unanswered for me.

Added some more tests and details in the README, hopefully the current state is enough for MVP @SergioBenitez

README.md
```diff
@@ -364,3 +364,34 @@ Escapes a string's HTML. Specifically, it makes these replacements:
...
+
+```jinja2
+{{ macros::input(label="Name", type="text") }}
+```
+Do note that macros, like other functions in Tera, require keyword arguments.
+If you are trying to call a macro defined in the same file, you will need to use the `{{ self::my_macro() }}`` syntax to call it the file its defined in.
+
+Macros can be called recursively but there is no limit to recursion so make sure you macro ends.
+
+There are a few restrictions to the content inside macros: no macros definitions, includes or blocks.
```
SergioBenitez: Why can't you do an `include` inside of a macro? That seems useful.

src/parser.rs
SergioBenitez: Right, but why can't there be a default name, like, the name of the file without the extension(s)?

src/render.rs
```diff
@@ -345,6 +363,63 @@ impl<'a> Renderer<'a> {
...
+            let real_namespace = match namespace.as_ref() {
+                "self" => {
+                    // TODO: handle error if we don't have a namespace
+                    self.macro_namespaces
+                        .last()
+                        .expect("Open an issue with a template sample please (mention `self namespace macro`)!")
+                        .to_string()
+                },
+                _ => {
+                    // TODO: String doesn't have Copy trait, can we avoid double cloning?
```
SergioBenitez: Yes. Keep a `Vec<&str>` instead. Or, a `Vec<Rc<String>>`, if you wish.

src/render.rs
```diff
@@ -345,6 +363,63 @@ impl<'a> Renderer<'a> {
...
 
+    fn render_macro(&mut self, call_node: Node) -> TeraResult<String> {
+        if let MacroCall {namespace, name: macro_name, params: call_params} = call_node {
+            // To find the real namespace we just look at the last one we got.
+            // If we have self, we take the last one otherwise we push it into the vec of namespaces
+            let real_namespace = match namespace.as_ref() {
+                "self" => {
+                    // TODO: handle error if we don't have a namespace
+                    self.macro_namespaces
+                        .last()
```
SergioBenitez: The logic here is somewhat unsettling. It's not clear to me how `macro_namespaces` gets populated the firt time. IE: How does this _not_ panic when calling a recursive macro? In that case the namespace will be `self`, but won't `macro_namespaces` be blank? If not, how does it get its elements?

src/render.rs
```diff
@@ -345,6 +363,63 @@ impl<'a> Renderer<'a> {
...
+                        .expect("Open an issue with a template sample please (mention `self namespace macro`)!")
+                        .to_string()
+                },
+                _ => {
+                    // TODO: String doesn't have Copy trait, can we avoid double cloning?
+                    self.macro_namespaces.push(namespace.clone());
+                    namespace.clone()
```
SergioBenitez: You don't need to `clone` here at all. Just use `.as_str` to get a pointer to the `String`.

src/render.rs
```diff
@@ -345,6 +363,63 @@ impl<'a> Renderer<'a> {
         Ok(output.trim_right().to_string())
     }
 
+    fn render_macro(&mut self, call_node: Node) -> TeraResult<String> {
+        if let MacroCall {namespace, name: macro_name, params: call_params} = call_node {
+            // To find the real namespace we just look at the last one we got.
+            // If we have self, we take the last one otherwise we push it into the vec of namespaces
+            let real_namespace = match namespace.as_ref() {
```
SergioBenitez: Use `as_str` or `&*namespace`.

src/render.rs
```diff
@@ -345,6 +363,63 @@ impl<'a> Renderer<'a> {
...
+                },
+                _ => {
+                    // TODO: String doesn't have Copy trait, can we avoid double cloning?
+                    self.macro_namespaces.push(namespace.clone());
+                    namespace.clone()
+                }
+            };
+
+            if let Some(Macro {body, params, ..}) = self.macros.get(&real_namespace)
+                .and_then(|m| m.get(&macro_name).cloned()) {
```
SergioBenitez: Better to bind the right-hand side of this match and then try to the match:

```
let x = self.macro.get(...).and_then(...);
if let Some(...) = x ...
```

src/render.rs
```diff
@@ -345,6 +363,63 @@ impl<'a> Renderer<'a> {
...
+                // fail early if we don't have the same number of args
+                let expected_params = params.iter().cloned().collect::<Vec<String>>();
```
SergioBenitez: Why is there so much cloning going on here? And why the `collect`? It's remarkably inefficient. If you just want to get a count, use `count`.

src/render.rs
```diff
@@ -345,6 +363,63 @@ impl<'a> Renderer<'a> {
...
+                        return Err(MacroCallWrongArgs(macro_name, expected_params, params_seen));
+                    }
+                    context.insert(param_name.to_string(), try!(self.eval_expression(exp)));
+                }
+                self.macro_context = Some(to_value(&context));
+                let mut output = String::new();
+                for node in body.get_children() {
+                    output.push_str(&try!(self.render_node(node)));
+                }
+                // Remove the namespace before returning if necessary
```
SergioBenitez: This comment doesn't help me understand why this is necessary or important or correct.

src/render.rs
```diff
@@ -345,6 +363,63 @@ impl<'a> Renderer<'a> {
...
+                    self.macro_namespaces.pop();
+                }
+                self.macro_context = None;
+                return Ok(output.trim().to_string());
+            } else {
+                return Err(MacroNotFound(real_namespace, macro_name));
+            }
+
+        } else {
+            unreachable!()
```
SergioBenitez: Tell me why: `unreachable!("This is why.")`.

Keats:
For 1, there is a test using a factorial macro, do you mean something else?

SergioBenitez:
No, that's perfect. Sorry, I missed it.

README.md
Keats: I was wondering about that and hoped someone would comment on it! I was not sure myself, I'll add it back

Keats:
How would you do the macro namespace logic? The current one makes sense to me so I'm interested in seeing other methods.

> Right, but why can't there be a default name, like, the name of the file without the extension(s)?
I can't reply to this inline for some reason.  It's an interesting idea I didn't consider before but that is too implicit imo

SergioBenitez:
I don't actually understand the namespace logic in your code, though I haven't spent much time looking at it. Can you explain what your code is doing?

I don't think that's too implicit. Almost every language with imports has a default name when the import isn't qualified, and I don't see why it should be any different in Tera.

README.md
SergioBenitez: Maybe add a test to ensure it works as expected?

Keats: There is one that try to define a macro inside a macro (in parser-failure)

SergioBenitez: Sorry, I meant there should be a test that checks that `include` works inside of a macro. That is, if a macro is called that uses `include`, the `include`d content is rendered.

Keats:
Fixed a couple of bugs and added some more comments in the code to explain it, hopefully it's clearer now - I can't find a simpler way myself.

If you have some time i'd love to see how you would do it to remove the allocations

src/render.rs
Keats: Didn't have to deal with lifetime too much yet, but how would you keep the `&str` values alive longer than the `if let` (without using `Rc`)? I tried a couple of approaches but I can't get a lifetime that lives long enough

src/render.rs
```diff
@@ -345,6 +364,81 @@ impl<'a> Renderer<'a> {
...
+            };
+
+            // We get our macro definition using the namespace name we just got
+            let macro_definition = self.macros
+                .get(&active_namespace)
+                .and_then(|m| m.get(&macro_name).cloned());
+
+            if let Some(Macro {body, params, ..}) = macro_definition {
+                let expected_params = params.iter().cloned().collect::<Vec<String>>();
+                let params_seen = call_params.keys().cloned().collect::<Vec<String>>();
```
Keats: I can remove them from there but then if i only collect the args on errors, I run into that error:

```rust
                for (param_name, exp) in call_params.clone() {
                    if !params.contains(&param_name) {
                        let expected_params = params.iter().cloned().collect::<Vec<String>>();
                        let params_seen = call_params.keys().cloned().collect::<Vec<String>>();
                        return Err(MacroCallWrongArgs(macro_name, expected_params, params_seen));
                    }
                    context.insert(param_name.to_string(), try!(self.eval_expression(exp)));
                }
```
Is there a way to iterate on a `HashMap` without moving the value and without `clone()`? I need the args name for the error message

Keats:
Pushed a commit with fewer allocations but still one `clone` for the one mentioned in a comment

src/render.rs
```diff
@@ -345,6 +364,81 @@ impl<'a> Renderer<'a> {
...
+                // fail fast if the number of args don't match
+                if params.len() != call_params.len() {
+                    let params_seen = call_params.keys().cloned().collect::<Vec<String>>();
+                    return Err(MacroCallWrongArgs(macro_name, params, params_seen));
+                }
+
+                // We need to make a new context for the macro from the arguments given
+                // Return an error if we get some unknown params
+                let mut context = HashMap::new();
+                for (param_name, exp) in call_params.clone() {
```
SergioBenitez: I'm pretty sure you don't need this `clone`. Just put an `&` before the `call_params`.

SergioBenitez: Ah, that's a bummer. All of the `eval` expression should allow references as well. Like taking in a T: Borrow<parser::Node> instead of just a parser::Node. But, meh.

SergioBenitez:
This is okay for now. I think we should work on removing many of those `clone` and `cloned` calls. They're unnecessary, for the most part.

Keats:
> This is okay for now. I think we should work on removing many of those clone and cloned calls. They're unnecessary, for the most part.

Tera is getting close to fully featured in my mind (once macro/escape/super are in) so the focus is going to be on things like that soon!

PR #71, opened 2016-11-14
HTML escaping

Keats:


src/utils.rs
```diff
@@ -0,0 +1,21 @@
...
+            '&' => output.push_str("&amp;"),
+            '"' => output.push_str("&quot;"),
+            '\'' => output.push_str("&#39;"),
+            '`' => output.push_str("&#96;"),
+            '/' => output.push_str("&#x2F;"),
+            _ => output.push(*c as char)
+        }
+    }
+
+    output.shrink_to_fit();
```
SergioBenitez: This is entirely unnecessary and will likely cause another allocation.

SergioBenitez:
Let's make this the default for strings, regardless of whether we decide to use `{{{ }}}` or some version of `{{ str | safe }}`.

Keats:
Default for string only for HTML files

SergioBenitez:
Right, that sounds fine: if the extension is "htm" or "html".

src/utils.rs
Keats: Why entirely unnecessary? we probably allocated too much initially.
I should set up a gitter or something so it's easier to discuss

SergioBenitez: Yeah, it's entirely likely that too much was allocated. But, that's okay: it's at most twice the original string size.

Keats: but it could be a big chunk, I think that would be an interesting thing to benchmark

Keats:
Added the autoescape mechanism.
Where did you get the ` and \ escaping in OWASP? From the site:
```
Escape the following characters with HTML entity encoding to prevent switching
into any execution context, such as script, style, or event handlers. Using
hex entities is recommended in the spec. In addition to the 5 characters
significant in XML (&, <, >, ", '), the forward slash is included as it helps
to end an HTML entity.

 & --> &amp;
 < --> &lt;
 > --> &gt;
 " --> &quot;
 ' --> &#x27;     &apos; is not recommended
 / --> &#x2F;     forward slash is included as it helps end an HTML entity
```

SergioBenitez:
The `\` escaping isn't necessary. The \` escape is necessary: older IE browsers treat it like a regular `'`. See https://github.com/OWASP/owasp-java-encoder/wiki/Grave-Accent-Issue for more information.

src/render.rs
```diff
@@ -53,15 +54,23 @@ pub struct Renderer<'a> {
     context: Value,
     tera: &'a Tera,
     for_loops: Vec<ForLoop>,
+    should_escape: bool,
 }
 
 impl<'a> Renderer<'a> {
     pub fn new(tpl: &'a Template, tera: &'a Tera, context: Value) -> Renderer<'a> {
+        let mut should_escape = false;
+        for ext in tera.autoescape_extensions.clone() {
```
SergioBenitez: Why is there a `clone` here? Should be able to simply do:

`for ext in &tera.autoescape_extenions { ... }`

src/render.rs
```diff
@@ -113,6 +122,17 @@ impl<'a> Renderer<'a> {
         match *node {
             Identifier { ref name, ref filters } => {
                 let mut value = try!(self.lookup_variable(name));
+
+                // Escaping strings if wanted for that template
+                if self.should_escape {
+                    match value {
+                      Value::String(s) =>  {
```
SergioBenitez: Don't match. `if let Value::String(s) = value { value = ... }`

src/tera.rs
```diff
@@ -19,6 +19,9 @@ pub struct Tera {
     pub templates: HashMap<String, Template>,
     pub filters: HashMap<String, FilterFn>,
     pub testers: HashMap<String, TesterFn>,
+    // Which extensions does Tera automatically autoescape on.
+    // Defaults to ["html", "htm", "xml"]
+    pub autoescape_extensions: Vec<String>,
```
SergioBenitez: Why is this a `Vec<String>` and not just a `Vec<&'static str>`? In fact, does it need to be a `Vec` at all? Why not a `[&'static str; 3]`? You'd need a `Vec` if you want to allow users to add their own extensions.

src/tera.rs
```diff
@@ -54,6 +57,7 @@ impl Tera {
             templates: templates,
             filters: HashMap::new(),
             testers: HashMap::new(),
+            autoescape_extensions: vec!["html".to_string(), "html".to_string(), "xml".to_string()]
```
SergioBenitez: No need to use a `String` here.

src/tera.rs
```diff
@@ -150,6 +158,12 @@ impl Tera {
         self.register_tester("string", testers::string);
         self.register_tester("number", testers::number);
     }
+
+    /// Select which extension(s) to automatically do HTML escaping on.
+    /// Pass an empty vec to completely disable autoescape
+    pub fn autoescape_on(&mut self, extensions: Vec<String>) {
```
SergioBenitez: `Vec<&'static str>`

src/tera.rs
```diff
@@ -158,6 +172,7 @@ impl Default for Tera {
             templates: HashMap::new(),
             filters: HashMap::new(),
             testers: HashMap::new(),
+            autoescape_extensions: vec!["html".to_string(), "html".to_string(), "xml".to_string()]
```
SergioBenitez: Same.

src/utils.rs
SergioBenitez: The `String` isn't likely to be stored anywhere, so the memory will be deallocated soon anyways. But `escape_html` will be called many, many times, and so its performance is very important.

src/render.rs
Keats: Hmm, I was sure I tried that before...guess not

Keats:
@SergioBenitez Updated

src/tera.rs
```diff
@@ -54,6 +57,7 @@ impl Tera {
             templates: templates,
             filters: HashMap::new(),
             testers: HashMap::new(),
+            autoescape_extensions: vec!["html", "html", "xml"]
```
SergioBenitez: Should one of these be `htm`?

src/render.rs
```diff
@@ -53,15 +54,23 @@ pub struct Renderer<'a> {
     context: Value,
     tera: &'a Tera,
     for_loops: Vec<ForLoop>,
+    should_escape: bool,
 }
 
 impl<'a> Renderer<'a> {
     pub fn new(tpl: &'a Template, tera: &'a Tera, context: Value) -> Renderer<'a> {
+        let mut should_escape = false;
+        for ext in &tera.autoescape_extensions {
```
SergioBenitez: Can probably make this a one-liner. Something like:

```
let should_escape = tera.autoescape_extensions.any(|ext| tpl.name.ends_with(ext));
```

src/render.rs
```diff
@@ -53,15 +54,23 @@ pub struct Renderer<'a> {
...
+            if tpl.name.ends_with(ext) {
+                should_escape = true;
```
SergioBenitez: `break` after this if not doing it with the one-liner.

src/tera.rs
```diff
@@ -158,6 +172,7 @@ impl Default for Tera {
             templates: HashMap::new(),
             filters: HashMap::new(),
             testers: HashMap::new(),
+            autoescape_extensions: vec!["html", "html", "xml"]
```
SergioBenitez: Same: one of this should be `htm`, probably.

src/render.rs
```diff
@@ -53,15 +54,23 @@ pub struct Renderer<'a> {
...
     tera: &'a Tera,
     for_loops: Vec<ForLoop>,
+    should_escape: bool,
 }
 
 impl<'a> Renderer<'a> {
     pub fn new(tpl: &'a Template, tera: &'a Tera, context: Value) -> Renderer<'a> {
+        let mut should_escape = false;
+        for ext in &tera.autoescape_extensions {
+            if tpl.name.ends_with(ext) {
```
SergioBenitez: `ends_with(ext)` will also match things like "file.not_html" or "this_is_my_html". Probably not intended. Should use something like `Path::new(&tpl.name).extension().map_or(false, |e| e == ext)`.

src/render.rs
```diff
@@ -113,6 +122,14 @@ impl<'a> Renderer<'a> {
         match *node {
             Identifier { ref name, ref filters } => {
                 let mut value = try!(self.lookup_variable(name));
+
+                // Escaping strings if wanted for that template
+                if self.should_escape {
+                    if let Value::String(s) = value {
```
SergioBenitez: Use `ref s` here to get rid of the `as_str` down there.

src/utils.rs
```diff
@@ -0,0 +1,35 @@
...
+/// / --> &#x2F;     forward slash is included as it helps end an HTML entity
+pub fn escape_html(input: &str) -> String {
+    let mut output = String::with_capacity(input.len() * 2);
+    for c in input.as_bytes() {
+        match *c as char {
+            '&' => output.push_str("&amp;"),
+            '<' => output.push_str("&lt;"),
+            '>' => output.push_str("&gt;"),
+            '"' => output.push_str("&quot;"),
+            '\'' => output.push_str("&#x27;"),
```
SergioBenitez: We don't need this one, right?

src/tera.rs
Keats: yep typo, thanks

src/render.rs
Keats: not doable, that would borrow the `value` and you can't assign to it anymore

SergioBenitez: Yeah, I suppose so. Ok.

src/utils.rs
SergioBenitez: Oh! Sorry. Didn't see the `'` there.

src/render.rs
Keats: Not too convinced by using Path::extension(), it only gets the bit after the last dot, which fails for files with several dots, ie if someone wants to escape `.php.html` or whatever. Tempted to just add the `.` for the default and ler user handle that themselves

SergioBenitez: Your example would work just fine: the thing after the final dot is `html`.

Keats: But users might add the extensions as `.php.html` in the Tera autoescape

src/tera.rs
```diff
@@ -150,6 +158,12 @@ impl Tera {
         self.register_tester("string", testers::string);
         self.register_tester("number", testers::number);
     }
+
+    /// Select which extension(s) to automatically do HTML escaping on.
+    /// Pass an empty vec to completely disable autoescape
```
SergioBenitez: I would add an example here.

src/tera.rs
```diff
@@ -19,6 +19,9 @@ pub struct Tera {
     pub templates: HashMap<String, Template>,
     pub filters: HashMap<String, FilterFn>,
     pub testers: HashMap<String, TesterFn>,
+    // Which extensions does Tera automatically autoescape on.
+    // Defaults to ["html", "htm", "xml"]
```
SergioBenitez: Comment needs to be fixed.

tests/render_ok.rs
```diff
@@ -12,6 +12,7 @@ use common::{Product, Review, read_file};
 
 fn assert_template_ok(path: &str, others: Vec<&str>) {
     let mut tera = Tera::default();
+    tera.autoescape_on(vec!["html"]);
```
SergioBenitez: Should be ".html", I believe.

SergioBenitez: Since it worked, this means no template is actually testing this. Should probably add a template that needs escaping.

Keats: I added a simple test for that in `render.rs`

Keats:
Asked a friend about `{{{ }}}` and `{{ | safe }}` and he raised what I think is a good point: the triple brackets are easy to miss compared to `{{ }}` while `| safe` is more in your face and I agree with that

Keats:
Also an interesting point is at which point does the escaping happens with filters:

- when we get the variable?
- when we push the output?

If we escape when we get the filters, then the filters will work on escaped content, which might not be what you want. On the other hand, it's quite rare to want to chain filters on a string marked ` | safe`.
I pushed a small patch to see how that'd look if putting `| safe` anywhere in the filter chain mark it as safe and won't be escaped

Keats:
@SergioBenitez ping

Keats:
Merging for now, we can look at it again later

PR #73, opened 2016-11-15
Add support for super() and multiple levels of extends

yonran:
Here is an implementation of `{{super()}}` and multiple levels of `{% extends "base" %}`. Probably it will conflict with the macros branch so feel free to ignore or not use it.

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -49,6 +49,7 @@ impl ForLoop {
 
 #[derive(Debug)]
 pub struct Renderer<'a> {
+    templates: Vec<&'a Template>,
     template: &'a Template,
     context: Value,
     tera: &'a Tera,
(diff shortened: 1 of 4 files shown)


Keats:
Thanks for that, I'll have a look once macros are in

src/tera.rs
```diff
@@ -78,11 +78,26 @@ impl Tera {
...
         }
 
-        let template = try!(self.get_template(template_name));
-        let mut renderer = Renderer::new(template, self, value);
+        let templates = try!(self.get_template_and_ancestors(template_name));
+        let mut renderer = Renderer::new(templates, self, value);
         renderer.render()
     }
 
+    fn get_template_and_ancestors(&self, template_name: &str) -> TeraResult<Vec<&Template>> {
```
Keats: Can we build that list of dependencies at compile time? I.e. store ancestors in the `Template` struct.
We have `parent` already, we could fill `Template::ancestors` once all templates are parsed and that moves the `CircularExtends` error at compile time too.

src/render.rs
```diff
@@ -373,32 +379,48 @@ impl<'a> Renderer<'a> {
...
+            Super => {
+                if self.block_definitions.is_empty() {
+                    Err(SuperOutsideBlock(self.templates.last().unwrap().name.clone()))
+                } else if self.block_stack_idx == 0 {
+                    let template_name = self.block_definitions[self.block_stack_idx].0.name.clone();
+                    let block_name = match &self.block_definitions[self.block_stack_idx].1 {
+                        &Block {ref name, ..} => name.clone(),
+                        _ => unreachable!()
+                    };
+                    Err(SuperNoParent(template_name, block_name))
```
Keats: that can be moved at parsing time as well, when we insert the blocks in `Template`, we already know if we have a parent or not

src/render.rs
```diff
@@ -373,32 +379,48 @@ impl<'a> Renderer<'a> {
...
-                            Block {body, ..} => {
-                                self.render_node(*body.clone())
-                            },
-                            _ => unreachable!()
-                        }
-                    },
-                    None => {
-                        self.render_node(*body)
-                    }
+                self.block_definitions.push((self.templates.last().unwrap(), Block {name: name.clone(), body: body.clone()}));
```
Keats: you could clone the node instead of creating a new `Block` no?

Keats:
Closing this in favour of #75 that also adds nested blocks and some more extend cases. Probably a bit buggy and needs more tests.

Thanks for the PR though, that was the base of #75.

yonran:
Sorry for taking too long, and thank you for pressing onward.

Keats:
No worries!

PR #75, opened 2016-11-27
Multiple level extends + super() + nested blocks

Keats:


Keats:
@yonran started on the super/extends while pushing most of the errors checks to the initialization of Tera, still missing the rendering though

Issue #79, opened 2016-12-06
Use 64 bits for int/float


Keats:
Right now it's 32

Should be done on the v0.5 branch

PR #85, opened 2016-12-08
Use 64 bits for int/float

terahxluna:
Closes https://github.com/Keats/tera/issues/79

Saw your your call for participation in This Week in Rust 159, here's my contribution =)

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -182,7 +182,7 @@ impl<'a> Renderer<'a> {
         }
     }
 
-    fn eval_math(&self, node: &Node) -> Result<f32> {
+    fn eval_math(&self, node: &Node) -> Result<f64> {
         match *node {
             Identifier { ref name, .. } => {
                 self.eval_ident(node)?
@@ -193,7 +193,7 @@ impl<'a> Renderer<'a> {
                         ).into()
                     ))
             },
-            Int(s) => Ok(s as f32),
+            Int(s) => Ok(s as f64),
             Float(s) => Ok(s),
             Math { ref lhs, ref rhs, ref operator } => {
                 let l = self.eval_math(lhs)?;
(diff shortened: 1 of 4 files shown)


Keats:
Thanks for that! 
i just pushed a big change to `0.5` branch so it needs to be rebased.

Would you be able to add a test in `render.rs` for math?

Something like:
```rust
#[test]
fn test_can_do_math() {
        let tests = vec![
            ("{{ 1 + 1 }}", "2"),
            ("{{ 1 + 1.1 }}", "2.1"),
            ("{{ 10 / 5 }}", "2"),
            // etc
        ];
        for (input, expected) in tests {
            tera.add_template("hey", input);
            assert_eq!(tera.render("hey", Context::new()), expected);
        }
}
```
Basically try to break the rendering using whatevet maths available

@SergioBenitez also left a comment in `render.rs` that applies here I believe:
```rust
// Since Tera only support 32 bit integers, this
// actually preserves all of the precision. If Tera
// switches to 64-bit values, use std::f32::EPSILON
// to get an approximation as before.
lhs_val = Value::F64(lhs_val.as_f64().unwrap());
rhs_val = Value::F64(rhs_val.as_f64().unwrap());
```

terahxluna:
I rebased it on v0.5 and added some math tests. I'll take a look at @SergioBenitez 's comment tonight.

src/render.rs
```diff
@@ -638,8 +638,13 @@ mod tests {
...
     #[test]
     fn test_render_math() {
-        let result = render_template("This is {{ 2000 + 16 }}.", Context::new());
-        assert_eq!(result.unwrap(), "This is 2016.".to_owned());
+        assert_eq!(render_template("{{ 1 + 1 }}", Context::new()).unwrap(), "2".to_owned());
+        assert_eq!(render_template("{{ 1 + 1.1 }}", Context::new()).unwrap(), "2.1".to_owned());
+        assert_eq!(render_template("{{ 3 - 1 }}", Context::new()).unwrap(), "2".to_owned());
+        assert_eq!(render_template("{{ 3 - 1.1 }}", Context::new()).unwrap(), "1.9".to_owned());
+        assert_eq!(render_template("{{ 2 * 5 }}", Context::new()).unwrap(), "10".to_owned());
+        assert_eq!(render_template("{{ 10 / 5 }}", Context::new()).unwrap(), "2".to_owned());
```
Keats: can you add `int * float` and `int / float` as well please?

Keats:
Looks good, thanks! Let me know when you are done with it

Keats:
@terahxluna ping

terahxluna:
Sorry for the delay.

I'm having some trouble to understand what would the use of std::f32::EPSILON do.

For what I understand, a conversion already takes place in lhs_val.as_f64(), so the loss of precision has already took place if the number is bigger than the fraction part of f64.

Keats:
Hmm i'll have a look later then, thanks!

PR #91, opened 2016-12-13
Add Context::from_btreemap 

erasin:
When I separate controllers and render the data in a template, use 'btreemap!' to create the data, because Context.data is private, so the additional function to populate the data.
 
   `fn from_btreemap(m: BTreeMap<String, Value>) -> Context`

like this

```
let data = btreemap!{
     "name".to_string() => to_value("konoha"),
      "age".to_string() =>  to_value(2)
};
render::new("tera.html", data) 
```

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -17,6 +17,24 @@ impl Context {
         }
     }
 
+    /// Use `btreemap!{}` create data to separation of data and rendering process.
+    ///
+    /// ```rust
+    /// #[macro_use] extern crate maplit;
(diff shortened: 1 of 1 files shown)


Keats:
Would the current `tera.value_render` work instead of using the context? (https://github.com/Keats/tera/blob/master/src/tera.rs#L143)

erasin:
thank you @Keats , I will try it.

Keats:
Let me know if you have any other issues!

PR #95, opened 2016-12-16
Fix try_get_value using inaccessible module

### LINKED ISSUE #94: try_get_value! refers to a private module
As I mentioned in [the reddit thread](https://www.reddit.com/r/rust/comments/5ijo5l/24_days_of_rust_tera/), the `try_get_value!` macro used in custom filters doesn't work outside the `tera` crate itself. This is due to this line in the macro expansion:

```rust
 return Err(::errors::TeraError::FilterIncorrectArgType(
```

but the `errors` module is private within `tera`. Here's a minimal example that fails to compile:
```rust
#[macro_use]
extern crate tera;
extern crate serde_json;

use std::collections::HashMap;
use tera::{TeraResult, Value};

pub fn dummy_filter(value: Value, _: HashMap<String, Value>) -> TeraResult<Value> {
    let _ = try_get_value!("dummy", "value", String, value);
    Ok(value)
}

fn main() {}
```

And the error is:
```
error[E0433]: failed to resolve. Maybe a missing `extern crate errors;`?
 --> src/tera.rs:9:13
  |
9 |     let _ = try_get_value!("dummy", "value", String, value);
  |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Maybe a missing `extern crate errors;`?
  |
  = note: this error originates in a macro outside of the current crate
```

zsiciarz:
Fixes #94.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -8,7 +8,7 @@ macro_rules! try_get_value {
         match ::serde_json::value::from_value::<$ty>($val.clone()) {
             Ok(s) => s,
             Err(_) => {
-                return Err(::errors::TeraError::FilterIncorrectArgType(
+                return Err(::tera::TeraError::FilterIncorrectArgType(
                     $filter_name.to_string(),
                     $var_name.to_string(),
                     $val,


zsiciarz:
Not sure about why Travis build failed... I ran a clean build of my fork and all tests pass without any build issues regarding `TeraError` not found.

Keats:
Don't worry about it, error types exporting is not great in 0.4, this is fixed in 0.5

PR #97, opened 2016-12-17
Fix include whitespace (#72)

### LINKED ISSUE #72: Fix include whitespace
Looking at `templates/included.html` we need to do weird things to keep indentation. Have a look and see if it's easily fixable

SeiryuZ:
This fix #72 

Cheers


Keats:
Thanks!

Keats:
It is actually harder than that if you want perfect whitespace handling but that's good enough I think!

PR #103, opened 2016-12-27
Added value_one_off function to render a single template with a Seria…

Alex-PK:
Hi, and sorry for the unrequested PR.

I noticed that there was a value_render function but not a value_one_off one, and I needed it for my project, so I implemented it.

Let me know if it needs changes.


Keats:
add a test to cover it and it's good to go!

Keats:
Thanks!

PR #105, opened 2016-12-30
Save a little clone

wdv4758h:
The improvement on my machine: (It's small, though)
perf


Keats:
Every little bit counts, thanks!

PR #106, opened 2016-12-30
Save a little clone, again

wdv4758h:
Don't know why the `tpl_name` in `Template::new`'s `find_blocks` function is unused.
Is it reserved for future use ? or it's something can be removed ?


Keats:
Ah I forgot to add it in the `bail!` call >_<, it's there to tell you at which file it encountered an error, I'll fix it

Keats:
Thanks again!

Keats:
Actually the template name will be given to the error in `Tera` so there's no need for that at all, I'll remove it

PR #107, opened 2017-01-03
Fixes and enhancements to the date filter

philwhineray:


src/tera.rs
```diff
@@ -350,7 +350,7 @@ impl Tera {
         self.register_filter("pluralize", number::pluralize);
         self.register_filter("round", number::round);
         self.register_filter("filesizeformat", number::filesizeformat);
-        self.register_filter("filesizeformat", number::date);
+        self.register_filter("date", number::date);
```
Keats: ah <_<

src/filters/number.rs
```diff
@@ -62,26 +62,43 @@ pub fn filesizeformat(value: Value, _: HashMap<String, Value>) -> Result<Value>
...
-/// `format` defaults to the ISO 8601 `YYYY-MM-DD` format
+/// Returns a formatted time according to the given `format` argument.
+/// `format` defaults to the ISO 8601 `YYYY-MM-DD` format.
+///
+/// Input can be an i64 timestamp (seconds since epoch) or an RFC3339 string
+/// (default serialization format for chrono::DateTime).
+///
 /// Time formatting syntax is inspired from strftime and a full reference is available
 /// on [chrono docs](https://lifthrasiir.github.io/rust-chrono/chrono/format/strftime/index.html)
 pub fn date(value: Value, args: HashMap<String, Value>) -> Result<Value> {
```
Keats: since that now works on both number and string, can you put it in `common.rs` and use pattern matching on the value like in that file?

Keats:
Thanks, facepalm for the registration of the date filter

philwhineray:
Will make the changes as requested and resubmit - it might take me a couple of days as I've quite a bit on.

If you want the registration fix immediately, feel free to cherry pick it or re-implement and I'll rebase.

philwhineray:
Hopefully this is what you had in mind

Keats:
Looks good, thanks!

PR #109, opened 2017-01-08
Fix date filter + clippy run

Keats:
@philwhineray I had to remove the `Local` use, otherwise the filter would try to be too clever and change the timezone (tests were not passsing on my machines for example)

Some work with timezone is needed but not sure how to tackle that yet

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ error-chain = "0.7"
 # used in slugify filter
 slug = "0.1"
 # used in striptags filters
-regex = "0.1"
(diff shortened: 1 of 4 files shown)


philwhineray:
Sorry for the extra work. I happen to be in UTC so I guess I got away with
it by accident. The Travis servers must be configured that way too.

I was originally planning to add a tz parameter to work like the format
one, but the chrono crate seems to be lacking good helpers in that area, so
I put it off.

Keats:
No worries, I would have missed it as well if I were still living in London

PR #112, opened 2017-01-11
Add add_template_file{s} methods.

### LINKED ISSUE #111: `add_template` and `add_templates` are confusing
The `add_template` and `add_templates` are confusing and error-prone. First, the documentation isn't clear what the arguments are/should be. But I won't go into this as I think the methods should be completely different.

I propose that the `add_template` method have the following signature:

```rust
fn add_template<P: AsRef<Path>>(&mut self, name: &str, path: P) -> Result<()>;
```

where `path` is the full path to the file. Tera should read the file, just like it does in `new`.

So, why this API? First, because reading a template file is not something the client should typically be doing. Maybe a different method, `add_raw_template` can be added to handle this case. But the default should be to read from a file. Second, the present API essentially forces you to use the filename as `name`. If it is not used as `name`, autoescaping isn't done since Tera doesn't know the path. Autoescaping should depend on the filename, not on the template's name.

Accordingly, the `add_templates` function should have the following signature:

```rust
fn add_templates<P: AsRef<Path>>(&mut self, templates: &[(&str, P)]) -> Result<()>
```

SergioBenitez:
Resolves #111.

### DIFF
--- a/benches/tera.rs
+++ b/benches/tera.rs
@@ -99,7 +99,7 @@ impl serde::Serialize for Product {
 
 #[bench]
 fn bench_parsing_basic_template(b: &mut test::Bencher) {
-    b.iter(|| Template::new("bench", SIMPLE_TEMPLATE));
+    b.iter(|| Template::new("bench", SIMPLE_TEMPLATE, None));
 }
 
 #[bench]
--- a/src/render.rs
+++ b/src/render.rs
@@ -71,10 +71,7 @@ pub struct Renderer<'a> {
 impl<'a> Renderer<'a> {
     pub fn new(tpl: &'a Template, tera: &'a Tera, context: Value) -> Renderer<'a> {
         let should_escape = tera.autoescape_extensions.iter()
-            .any(|ext| {
-                // This first part is only to deal with the one_off hack.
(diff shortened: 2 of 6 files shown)


SergioBenitez:
@Keats How are you feeling about this? I'm hoping to release the next version of Rocket soon, which requires this functionality.

Keats:
I just commented on #111 
If you can do a PR that only adds the adding template by file feature while leaving out the renaming, I'll merge it and then we can see what needs to be changed to support Rocket usecase

SergioBenitez:
The renaming of what? This doesn't rename anything.

Keats:
`fn add_template_file<P: AsRef<Path>>(&mut self, name: &str, path: P) ` should be `fn add_template_file<P: AsRef<Path>>(&mut self, path: P) `

SergioBenitez:
But that would defeat the purpose of this PR.

I _really_ don't understand your objection here. It seems arbitrary and baseless. Just about every template engine allows you to name templates any way you'd like. Why make Tera contrarian? Programming languages don't prohibit you from renaming modules. Why would you want to enjoin a particular name?

PR #116, opened 2017-01-21
Fix #115

### LINKED ISSUE #115: Don't use relative names in macro
`compile_templates` macro shouldn't require importing `tera::Tera` type, but use [$crate variable](https://doc.rust-lang.org/beta/book/macros.html#the-variable-crate) instead.

LFalch:
Had to re-export ::serde_json::value::from_value in crate root.
Tell me if you think that should be done in another way.


Keats:
Thanks!

PR #118, opened 2017-01-24
Add add_template_file{s}

Keats:


Keats:
@SergioBenitez have a look at that

SergioBenitez:
I don't...understand. It looks like you've taken my PR/code and simply made the path an `Option`? What's the point? The user already has the path, since they need to pass it in, so if they want to set the name to the path, it's trivial to do so. This just complicates the API and reasoning, in my mind.

If you want the path to be the canonical name, just say so in the docs: "It is encouraged to use the path to the template as the name of the template."

Keats:
Actually I wrote it offline and realised how similar it was only when I got online and compared it with your PR (and took the error message for file reading from it).

> Actually, I find this very odd; a template always has a name, but it may or may not have a path. It doesn't make sense to say that a name is optional.

I agree on the first part, which is why a new name it is an `Option`, by default it takes the name from the path. Encoding that in the types rather than in the docs as suggested is better. I find an API forcing me to do `add_template_file(path, path)` considerably worse. 
Either way this function is not going to be part of the public API (available but not shown in the docs) until extending Tera instances is fleshed out

Issue #124, opened 2017-01-29
Merge 2 Context Structs


sklopi:
For me it would be very useful, to be able to merge 2 Context structs, where already defined values get overwritten. If there is any interest in this feature i would like to implement this feature.

Keats:
Something like `Context::extend(&mut self, &Context)`? Go for it

sklopi:
`extend` might be better, because that's how it is called in various other (non-rust) libraries.
This would also represent that existing values can be overwritten.

PR #126, opened 2017-01-30
Add extend method to Context struct.

sklopi:
This method allows extending one context with the content of a second context.
The second context is dropped. 
References this issue: #124


Keats:
Thanks!

Issue #102, opened 2016-12-27
Support Global Functions


SergioBenitez:
Tera doesn't appear to support [filter sections](http://jinja.pocoo.org/docs/dev/templates/#id11) at the moment. It's an important feature, especially for things like form helpers. Consider supporting them.

SergioBenitez:
An alternative is to support global macros (in the global namespace) that can be added via a call to `Tera`.

Keats:
Do you have examples for form helpers? I don't think I've used those before. 
The one linked should be fairly simple to implement but those filters only apply to strings right?

SergioBenitez:
See the `form_tag` helper in [Rails](http://guides.rubyonrails.org/form_helpers.html).

Keats:
That looks like macros calls rather than filter sections indeed. I don't think I want to allow global macros as magically loading macros is too implicit for my tastes, I want to know where the macros are coming from

SergioBenitez:
In _every_ language there is a global namespace of some sort. There's no reason not have a global namespace for macros here, too. If the macro doesn't have a namespace, then it's coming from the global or local namespace. I don't think there's much magic in that reasoning.

Keats:
I agree with some kind of global namespace, like the one for filters. 
I'm not planning to make anything of the sort for macros though, I'll keep the issue open as actual filter sections might be a wanted feature

SergioBenitez:
That seems arbitrary and inconsistent. Why do filters get a global namespace but not macros? What's the fundamental difference between them that justifies this?

Keats:
I don't want macros to be global since that would be terrible in terms of UX in my opinion. 

You're extending something like `flask-admin` and create a macro called `input` for your own code. What happens if the `admin` template were also defining an `input` macro that was used everywhere, which one gets used? You'd still need some kind of namespace to resolve that.
The same could happen with filters but it's much less likely that a filter like `length` gets redefined vs creating a `input`, `button` macro.

Overall, macros are about being DRY and explicit (vs what other template engine would do and just `include` another template) and making them global would go against that (again, imo).

SergioBenitez:
[...]

PR #128, opened 2017-02-01
Implemented filter sections

dswd:
This implements the filter sections from #102.
According to the tests it seems to work fine. However I am pretty unsure about the parser modifications as I wrote this part mostly by trial and error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -523,6 +523,32 @@ Escapes a string's HTML. Specifically, it makes these replacements:
 - `` ` `` is converted to `&#96;`
 
 
+### Filter sections
+Whole sections can also be processed by filters if they are encapsulated in `{% filter name %}` and `{% endfilter %}`
+tags where `name` is the name of the filter.
+
+Example:
+```jinja2
+{% filter upper %}
+    Hello
+{% endfilter %}
+```
+
+This example transforms the text `Hello` in all upper-case (`HELLO`).
+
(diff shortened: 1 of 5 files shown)


Keats:
Looks good!

Can you add a section in the README about it and a failing test in `tests/render_fails.rs` for something like:
```jinja2
{% filter round %}
hello
{% endfilter %}
```

Keats:
Thanks!

SergioBenitez:
This doesn't work correctly. Anything after an `endfilter` tag is omitted from the rendered template. For instance, if we have the following template:

```jinja
<!DOCTYPE html>
<html>
   <head>
      <meta charset="utf-8" />
      <meta name="viewport" content="width=device-width" />
      <title>Tera Filter Section</title>
   </head>
   <body>
      {% filter my_filter %}
         Hello.
      {% endfilter %}
   </body>
</html>
```

With the following helper:

```rust
fn helper(value: Value, map: HashMap<String, Value>) -> tera::Result<Value> {
    Ok(value)
}
```

This renders:

```html
<!DOCTYPE html>
<html>
   <head>
      <meta charset="utf-8" />
      <meta name="viewport" content="width=device-width" />
      <title>Tera Filter Section</title>
   </head>
   <body>

         Hello.
```

Keats:
Ah damn... I'll fix that in the next release

Keats:
It's fixed in master, i'll release along with a couple of other things next week

PR #131, opened 2017-02-04
Fix path prefix trimming on Windows.

SergioBenitez:
Looks like template naming has been broken on Windows for quite a while via the `Tera::new(...)` function. This change resolves the issue. See https://github.com/SergioBenitez/Rocket/issues/169#issuecomment-277435725 for a full explanation of the problem. Please roll this change out to previous versions of Tera as well.


Keats:
Thanks, It's in 0.7.1
I'm not planning to backport that though, Tera is still too much in flux (ie pre v1) to support more than just the current version

SergioBenitez:
Can you backport to `0.6`? It's a single release, and one rather important change.

PR #133, opened 2017-02-06
Update chrono.

Eijebong:
Since nothing seems to break in tera between 0.2 and 0.3, I just updated chrono to allow both version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,7 @@ url = "1"
 # used in filesizeformat filter
 humansize = "1"
 # used in date format filter
-chrono = "0.2"
+chrono = "0.3"
 
 [dev-dependencies]
 serde_derive = "0.9"


Cargo.toml
```diff
@@ -27,7 +27,7 @@ url = "1"
 # used in filesizeformat filter
 humansize = "1"
 # used in date format filter
-chrono = "0.2"
+chrono = ">=0.2, < 0.4"
```
Keats: any reason it can't be `0.3`?

LFalch: It can

Keats: Make it  `chrono = "0.3"` please and I'll merge!

Eijebong: Ok, I was just allowing chrono 0.2 to be used with tera, but I'm ok with 0.3 only ^^

Keats:
Thanks!

PR #138, opened 2017-03-01
Fix usage of join filter in documentation

mre:
I've tested the Django syntax as described in the documentation, but it gives me an "invalid syntax" error. Therefore I've changed it to the syntax of all other filters and that works.


Keats:
Too much copy/paste, thanks!

PR #140, opened 2017-03-04
Readability ideas

Peternator7:
I was working on changing how the strings are handled like I previously mentioned, and found a few places where I think the code can be refactored to improve readability. All the tests still pass. Let me know what you think!

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -451,7 +451,7 @@ impl_rdp! {
                 Ok(Some(Node::VariableBlock(Box::new(exp?))))
             },
             (_: raw_tag, &body: raw_text, _: endraw_tag) => {
-                Ok(Some(Node::Raw(body.to_string())))
+                Ok(Some(Node::Raw(body.trim().to_string())))
             },
             (_: filter_tag, _call: fn_call, &name: simple_ident, params: _fn_args(), body: _template()) => {
                 Ok(Some(Node::FilterSection{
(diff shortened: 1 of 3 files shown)


src/parser.rs
```diff
@@ -451,7 +451,7 @@ impl_rdp! {
                 Ok(Some(Node::VariableBlock(Box::new(exp?))))
             },
             (_: raw_tag, &body: raw_text, _: endraw_tag) => {
-                Ok(Some(Node::Raw(body.to_string())))
+                Ok(Some(Node::Raw(body.trim().to_string())))
```
Keats: it would be better to keep the trimming in the renderer so we only have one place to look at

src/render.rs
```diff
@@ -441,29 +439,17 @@ impl<'a> Renderer<'a> {
...
-                }
-                i += 1;
-            }
-            output = output.trim_right().to_string();
-        } else {
-            self.for_loops.pop();
+        for _ in 0..length {
+            output.push_str(self.render_node(body)?.trim_left());
+            // Safe unwrap
+            self.for_loops.last_mut().unwrap().increment();
```
Keats: ❤️

Keats:
Looks very good minus the `trim` in the parser!

Peternator7:
Good! I undid the trim changes

Keats:
Do you have more coming or can I merge that?

Peternator7:
I have 1 more idea; it's a simple, but larger refactoring. The idea was to reduce the rightward drift that match statements can cause. Take a look at this PR I put on myself. If you like it, I'll merge it then you can complete this PR. Otherwise feel free to merge this one as is.

https://github.com/Peternator7/tera/pull/1

Peternator7:
This is ready for merging!

Keats:
Thanks a lot!

PR #141, opened 2017-03-05
Improved Macro performance

Peternator7:
I made a few very minor tweaks to the macro renderer that give nice speed ups rendering macros

Pre changes:
```
test bench_rendering_inheritance_and_macros    ... bench:      11,781 ns/iter (+/- 373)
test bench_rendering_only_macro_call           ... bench:       1,515 ns/iter (+/- 80)
```

Post Changes
```
test bench_rendering_inheritance_and_macros    ... bench:       9,168 ns/iter (+/- 1,280)
test bench_rendering_only_macro_call           ... bench:       1,450 ns/iter (+/- 137)
```

Using references was an easy fix. The other change is to use the `serde_json::map::Map` instead of `HashMap` when creating the context since it appears to be an almost free conversion from `Map` to `Value`.

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -1,5 +1,6 @@
 use std::collections::{VecDeque, HashMap};
 
+use serde_json::map::Map;
 use serde_json::to_string_pretty;
 use serde_json::value::{Value, to_value, Number};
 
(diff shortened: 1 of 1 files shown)


Keats:
Very nice, thanks!

PR #146, opened 2017-03-08
Reload + Extend

Keats:


Keats:
Need a test for the reloading from glob with and without extend but should be fine after that!

PR #152, opened 2017-04-04
feat: add base64 encode/decode filters

sebasmagri:
This implements `base64encode` and `base64decode` filters.

This is done mainly for the use case of static stite generators such as gutenberg. For this cases, pretty often you need to implement more dynamic features that rely on current data and not in data available at the moment of the html render. For example, being able to manipulate content in a posts index based on the comparison of the post's date and the current browser's date in the client. However, when giving data processing control to the client, you can end up having difficulties handling multiline content or custom data. In such cases it's possible to use base64 to encode the data in the render stage and have the browser decode the data when needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,8 @@ url = "1"
 humansize = "1"
 # used in date format filter
 chrono = "0.3"
+# used in the base64 format filter
+rustc-serialize = "*"
 
 [dev-dependencies]
 serde_derive = "0.9"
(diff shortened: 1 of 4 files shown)


Keats:
Thanks for that.
It seems that this belongs to gutenberg rather than Tera though, I don't think base64'ing things is that common in templates and deserve to be in it.

I'm planning to add a couple of filters in gutenberg today or tomorrow, I can add a base64 as well.

Closing this one and created https://github.com/Keats/gutenberg/issues/28 to track it in gutenberg

PR #159, opened 2017-04-09
Fix `one_off` example

matklad:
Hi! Looks like there's an argument missing in the example, and it is `,ignore` and dodges doctests :)


Keats:
Thanks!

Issue #162, opened 2017-04-20
Upgrade to Serde 1.0


SergioBenitez:
This is blocking https://github.com/SergioBenitez/Rocket/pull/272.

skariel:
@SergioBenitez just made a PR

Keats:
It's in 0.10.0!

PR #163, opened 2017-04-20
updated serde to v1

skariel:
see #162


Keats:
Thanks! I'll publish a new version (probably 0.10 as I re-export some Serde types so that would be a breaking change) after I test #164 a bit more

PR #165, opened 2017-04-24
Fixes handling loop variables

tailhook:
Previously if you had loop like `for x_y, x in z` the key variable `x_y`
was resolved same as `x` because it's name is a prefix of a value
variable. This commit fixes the bug.


Keats:
Thanks! I'll publish a new version later today

Keats:
It's in 0.10.1

PR #169, opened 2017-04-28
Global functions

Keats:
I need to try making some global functions in Gutenberg to make sure the design is ok

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.10.3 (unreleased)
+
+- Add global functions, see README
+
 ## 0.10.2 (2017-05-03)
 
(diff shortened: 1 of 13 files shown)


Keats:
@khorolets that should solve your `range` issue

@phil-opp that also contains the doc update for macros `self` and opens the way for the `get_page` function you mentioned on the gutenberg repo. That also opens the way to shortcodes like Twitter/Instagram in http://gohugo.io/extras/shortcodes/ where you can do a http request to an API from a template

@SergioBenitez does that fix @102 for you?

mackwic:
This is great ! 👍

102:
@Keats it was #102, not @102 🌝

Keats:
woops sorry!

Issue #168, opened 2017-04-26
How to access fields after filter?


phil-opp:
I'm trying to print the title of the last post in my `index.html`, but I can't find the correct syntax for this. Things I tried:

```jinja
{{ (pages | last).title }}
{{ pages | last .title }}
{{ {pages | last}.title }}
{{ pages.-1.title }}
{{ pages[-1].title }}
{% last_page = pages | last %}{{ last_page.title }}
{% set last_page = pages | last %}{{ last_page.title }}
```

I even tried to create a macro for this:

```jinja
macros::latest_post_link(page=pages | last)
macros::latest_post_link(page=(pages | last))
```

None of the above works. What's the correct way to do this?

Keats:
Hmm `macros::latest_post_link(page=pages | last)` should work in theory I think, I'll have a look.

I need to add the `{% set last_page = pages | last %}` construct to Tera when I have time. It should be pretty simple, the only issue being making sure you pop that value from the context correctly when it goes out of scope.

Keats:
So the issue with `macros::latest_post_link(page=pages | last)` was that I wasn't allowing whitespace in arguments for some reason. `macros::latest_post_link(page=pages|last)` was working.

I relaxed that in https://github.com/Keats/tera/commit/aca43d1e9c8882db0135fa510ac2a700feb3ae61 and nothing else seems to break so that should be in the next release

phil-opp:
Ok, the macro solution works without whitespace. Thanks!

> I need to add the {% set last_page = pages | last %} construct to Tera when I have time.

This would be great, since it would also allow to reverse the order of pages by doing `{% pages_sorted = pages | reverse %}` and the access the first/second page through `pages_sorted.0`/`pages_sorted.1`.

Keats:
`set` is in 0.10.4

PR #170, opened 2017-05-04
Add set tag

Keats:


Keats:
@phil-opp this should solve #168

PR #177, opened 2017-05-12
Fixes invalid dispatch of attributes after `set` in loop

tailhook:


tests/templates/loops.html
```diff
@@ -11,8 +11,8 @@
       {% if number_reviews > 10 or show_more %}
         <p>All reviews</p>
         {% for review in reviews %}
-          <h3>{{review.title}}</h3>
-          {% for paragraph in review.paragraphs %}
+          <h3>{{review.title}}</h3>{% set rev = review %}
```
Keats: can you create a new test for that? otherwise I'm never going to remember why it's there :)
If you can't, no worries i'll do it myself after mergin

Keats:
Thanks a lot!

tests/templates/loops.html
tailhook: I was trying to make a minimal change. Overall, yes this needs to be a separate test. And `lookup_variable` needs some refactoring now. But I'm not sure I'll have more time to spent on this.

Keats: no problem, thanks

PR #191, opened 2017-06-22
Bump chrono to 0.4

alex:


Keats:
Thanks!

alex:
@Keats would you be willing to do a release with this?

Keats:
Ah i didn't realise this was causing issues, will release a bit later today

PR #200, opened 2017-07-27
create a code block for the `escape_html` docs

Thomasdezeeuw:
Currently everything is printed on a single line. This change creates a code block that prints everything on it's own line.


Keats:
Thanks!

Issue #196, opened 2017-07-25
setting variables escapes them


Riamse:
```jinja
{% set line = "&" %}
{% set line2 = line %}
{% set line3 = line2 %}
<span>{{line3}}</span>
```
It'll print `&amp;amp;amp;` which gets rendered as "&amp;amp;amp;". I do not think this is intentional.

Riamse:
Seems that `{{line3|safe}}` gives us "&amp;amp;amp;", which is equivalent to `{{line2}}`, i.e. it undoes one level of escaping. You can't repeat safe pipes, though, it seems...

Riamse:
Seems that piping into safe doesn't actually undo escaping, it just prevents escaping from happening https://github.com/Keats/tera/blob/bc2ade03b2a3f1e44e8fc73816841237807cc289/src/render.rs#L275 .

Riamse:
Seems that the evaluation of Node::Set at https://github.com/Keats/tera/blob/bc2ade03b2a3f1e44e8fc73816841237807cc289/src/render.rs#L439 calls https://github.com/Keats/tera/blob/bc2ade03b2a3f1e44e8fc73816841237807cc289/src/render.rs#L330 which, when acting upon an ident, calls https://github.com/Keats/tera/blob/bc2ade03b2a3f1e44e8fc73816841237807cc289/src/render.rs#L244 which brings us back to the earlier link at https://github.com/Keats/tera/blob/bc2ade03b2a3f1e44e8fc73816841237807cc289/src/render.rs#L275 .

Calling `{% set x = y %}` shouldn't introduce the side effect of escaping `y`. I think a simple duct-tape fix would be to pipe a set assignment into the `safe` filter, but I think there might also be merit in having a separate function for this, or taking the `safe` pipe logic out of `eval_ident`.

What do you think?

Keats:
> You can't repeat safe pipes, though, it seems...
Yeah right now the `safe` is applied after all the filters which is not what people expect I guess. I don't really remember why I did it that way though... Changing that would be a breaking change thought so that will be for 0.11

For the second case, I agree that the `set` shouldn't escape `y` in that case. I'm not too sure how to fix it, maybe applying an unescape before setting the variable? I'll think about it but I don't think we want to take the `safe` logic out of `eval_ident`, it looks like it would be too easy to forget escaping in some cases otherwise.

Riamse:
[...]

PR #201, opened 2017-07-27
Fix issue #196, {% set %} shouldn't escape rvalues

Riamse:
A set call now pipes the rvalue into the `safe` filter before evaluating the expression. (Note that as only identifiers get escaped, we only have to do this to identifiers.)

### DIFF
--- a/src/render.rs
+++ b/src/render.rs
@@ -436,6 +436,12 @@ impl<'a> Renderer<'a> {
                 let val = match **value {
                     MacroCall {..} => to_value(self.render_macro(value)?).unwrap(),
                     GlobalFunctionCall { .. } => self.eval_global_fn(value)?,
+                    Identifier { ref name, ref filters } => {
+                        let mut filters: VecDeque<Node> = VecDeque::new();
+                        filters.push_front(Filter { name: "safe".to_owned(), params: HashMap::new() });
+                        self.eval_expression(&Identifier { name: name.to_owned(), filters: Some(filters) })?
+                    },
+
                     _ => self.eval_expression(value)?,
                 };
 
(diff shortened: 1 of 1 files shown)


src/render.rs
```diff
@@ -436,6 +436,12 @@ impl<'a> Renderer<'a> {
                 let val = match **value {
                     MacroCall {..} => to_value(self.render_macro(value)?).unwrap(),
                     GlobalFunctionCall { .. } => self.eval_global_fn(value)?,
+                    Identifier { ref name, ref filters } => {
+                        let mut filters: VecDeque<Node> = VecDeque::new();
+                        filters.push_front(Filter { name: "safe".to_owned(), params: HashMap::new() });
+                        self.eval_expression(&Identifier { name: name.to_owned(), filters: Some(filters) })?
```
Keats: can you add a comment explaining why this is done and a test for it?

Riamse: sure thing

Keats:
I just checked and this actually affects macro and filters arguments as well so this is probably not be the correct fix. Ie a recursive macro called like `{{ macros::label_for(obj=obj, sep="<!doctype html>") }}` will escape the doctype on every call.

I'm now thinking of the alternative you mentioned in #196: `taking the safe pipe logic out of eval_ident`.
The first step is probably to add tests for all cases (`set`, filter and macros) that would be currently failing and then fix it

Riamse:
That's unfortunate. How do you think the function will look in the future?

Keats:
I'm thinking of setting a flag on the renderer when it should NOT escape (set, arguments to filters/macro) that gets disabled once it has retrieved the ident

Keats:
Turns out the renderer already has such a flag so the result looks like #207 
Some copy&paste but it's not too bad imo

Keats:
Closing in favour of #207

Issue #202, opened 2017-07-27
Issue with recursing macro for recursive struct


kardeiz:
I'm trying to use a recursive macro on a recursive struct like so:

```jinja
{% macro formatted_label_for(subject, sep) %}
  {% if subject.parent %}
    {{ self::formatted_label_for(subject=subject.parent, sep=sep) }}
    {{sep}}
  {% endif %}
  {{subject.label}}
{% endmacro formatted_label_for %}
```

However, this crashes with:

```
thread '<unknown>' has overflowed its stack
fatal runtime error: stack overflow
```

On examination, it looks like the object itself (`subject`) is getting passed to the macro call, rather than its `parent`. I've also tried `set`ting the parent as its own variable before calling the macro, but that doesn't seem to make any difference.

Am I missing something or is this a known limitation?

Keats:
Hmm that looks like a bug, it should be working.. Can you open a PR with a test as a reproduction?

kardeiz:
I added a test that crashes (see PR #203 ).

Interestingly, this only happens when the macro is called inside a loop. I don't know what to make of this. Please let me know if I overlooked something. Thanks!

Keats:
So the issue is that the code that finds the variable will see there is a forloop and will try to get the value from there rather than the macro context, resulting in pretty much an infinite loop.
That seems to point out that the current context system is a bit broken as it doesn't take into account whether a forloop is in a macro or not. For now you can rename the macro param to something other than the forloop name (`obj` currently) while I figure out something.

I'll take any idea!

Keats:
I believe https://github.com/Keats/tera/pull/205 fixes it

kardeiz:
Wow, that's awesome! I would have been happy with the rename workaround as a decent short-term fix. Thanks!

Keats:
The fix is in 0.10.9

PR #203, opened 2017-07-28
Calling a recursive macro inside a loop can break things

kardeiz:
Adds a crashing test (sorry, I didn't dig too deeply into how tests were organized so I just put the test in `lib.rs`).

Test crashes with:
 ```
thread 'test_nested_object' has overflowed its stack
fatal runtime error: stack overflow
```

(Is it possible to capture SOs in cargo test?)

See [Issue 202](https://github.com/Keats/tera/issues/202)

### DIFF
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -31,3 +31,54 @@ fn test_can_load_template_files() {
     assert!(tera.get_template("tests/templates/basic.html").is_ok());
     assert!(tera.get_template("basic.html").is_ok());
 }
+
+
+#[test]
+fn test_nested_object() {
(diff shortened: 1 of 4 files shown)


Keats:
> (Is it possible to capture SOs in cargo test?)

Not as far as I'm aware. 

Thanks, I'll have a look! Tera also has another stack overflow due to the parser library that should be fixed in the next version of it but I don't think it's the same.

Keats:
Ok the issue seems that it just keeps on trying to call the macro (with the right variables) over and over and never stops, definitely a Tera issue

PR #214, opened 2017-08-31
Remove whitespace like Jinja2

Flakebi:
Hi,
thanks for this good working template engine.
I was using tera to generate some source code and it removed a bit too much whitespace.
Something like `{% if x %}pub {% endif %}struct X ...` gets merged to 'pubstruct' and I found no way to prevent this except inserting a newline in between.

For this reason I implement the whitespace removal of Jinja2 (trim_blocks and lstrip_blocks):
This means, a newline directly following a control tag (surrounded by `{% %}`) will be removed.
Also any indentation before a control tag will be removed.

In other words, if a line contains only a control tag (maybe with indentation), the line will be removed completely from the output.
All other whitespace will not be changed.

I needed to change the tests slightly to match the changed output (they did a good job and revealed some bugs in my code :) ).
If you want, I can add some tests specific to the whitespace behaviour.

I also thought about the additional whitespace control of Jinja2 so `{%- ...` removes all whitespace before the tag `{%+` preserves everything before.
They can also be combined like `{{+ "x" -}}`.
But I didn't find a nice way to store these markers in the nodes.


Keats:
Thanks for that!

> I also thought about the additional whitespace control of Jinja2 so {%- ... removes all whitespace before the tag {%+ preserves everything before.
> They can also be combined like {{+ "x" -}}.
> But I didn't find a nice way to store these markers in the nodes.

I've actually started working on tera 0.11 which has a complete rewrite of the parser and will support for `{%-` and `{%+`. 
Merging that would be a breaking change so I will include it in the 0.11 branch once I'm done with the parser!
I'll ping you once it's in a usable state.

Keats:
By the way we don't need to copy exactly the system Jinja2 uses, I find it a bit confusing myself. If you have any ideas on how you would improve it, I'll take them!

Keats:
Closing in favour of #218 which is almost done

PR #221, opened 2017-10-21
Fix doccomment for Tera::parse()

nabijaczleweli:
Compare: https://docs.rs/tera/0.10.10/tera/struct.Tera.html#method.parse


Keats:
thanks!

PR #218, opened 2017-09-24
V0.11

Keats:

PR #224, opened 2017-10-27
Add tester starting_with and ending_with

upsuper:


upsuper:
The failure doesn't seem to be related.

Keats:
Could you make the PR against the 0.11 branch instead? (https://github.com/Keats/tera/pull/218)

docs/content/docs/templates.md
```diff
@@ -575,6 +575,19 @@ Example:
 ### iterable
 Returns true if the given variable can be iterated over in Tera (ie is an array/tuple).
 
+### startswith
```
Keats: can we name it `starting_with` so it flows better with `is`?

docs/content/docs/templates.md
```diff
@@ -575,6 +575,19 @@ Example:
...
+Returns true if the given variable is a string starts with the arg given.
+
+Example:
+```jinja2
+{% if path is startswith "x/" %}
```
Keats: 0.11 require parentheses around tests params so that should be `{% if path is startswith("x/") %}

docs/content/docs/templates.md
```diff
@@ -575,6 +575,19 @@ Example:
...
+    In section x
+{% endif %}
+```
+
+### endswith
```
Keats: same as above, `ending_with`?

docs/content/docs/templates.md
upsuper: Ah, ok. Probably it makes sense to have a doc test for these documents as well :)

Keats:
Thanks!

PR #225, opened 2017-10-27
Add filter split

upsuper:


Keats:
Could you make the PR against the 0.11 branch instead? (#218)

upsuper:
Updated.

Keats:
Thanks!

PR #226, opened 2017-10-28
Add json_encode filter to print any value as a JSON string

Alex-PK:
Hi,

this PR adds a json_encode() filter, to transform any Value into a JSON string.

I'm using it often in my projects, so I thought it could be useful to have it the core.
I kept testing and docs at the minimum to see if there is interest, but I can add more if needed.

I am using Tera with escaping off, so I'm not sure how it will react. It will probably break any value containing some sort of HTML. :) But it's meant to be used inside js templates or in <script> tags, so escaping should be disabled anyway.

### DIFF
--- a/docs/content/docs/templates.md
+++ b/docs/content/docs/templates.md
@@ -544,6 +544,15 @@ Example: `{{ sections | get(key="posts/content") }}`
 Split a string into an array of strings, separated by a pattern given.
 Example: `{{ path | split(pat="/") }}`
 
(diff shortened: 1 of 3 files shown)


Keats:
Can you add it to the docs and do the PR on the v0.11 branch please? The code itself looks ok

Alex-PK:
Done. Thank you.

Keats:
Thanks!

PR #228, opened 2017-11-12
Remove Sized constraint on T in Context.add()

jturner314:
This makes it possible to call `context.add("foo", "bar")`.


Keats:
Thanks again

PR #230, opened 2017-11-18
WIP: tera contains

stephanbuys:
Basic support for substring matching. @Keats - am I on the right track? I couldn't get this to work in a template, how should it be used?

All the best!


Keats:
That should be a test, not a filter.
I have some time today so I'll add it directly in the 0.11 branch, thanks though!

Keats:
It's in https://github.com/Keats/tera/pull/218/commits/db07fe5f6979fec3a7011f97053619d6d6880393 forgot to link it there

Issue #236, opened 2017-11-27
Allow other type of quotes for strings in the parser


Keats:
Currently only double quotes are allowed, which is a bit annoying if you want to `set` a string which has double quotes for example. A simple solution is to allow simple quotes and backticks as synonyms for strings and treat all of them as the same.

This should be done on the 0.11 branch

PR #237, opened 2017-11-30
Allowed other type of quotes for strings #236

ragne:
Expanded the grammar with doublequotes and backticks and added them to the testcase.

I tried not to deviate from the description of the issue, but I'm not sure if it intended for `Rule::string` to _"decay"_ to the `Rule::text` when there is an unmatched quote(i.e `"\"string\""` is a `Rule::string`, but `"\"string"`(note the missing quote) is a `Rule::text`.

Note: it's a behaviour "inherited" from the `Rule::string` grammar.

### DIFF
--- a/src/parser/tests/lexer.rs
+++ b/src/parser/tests/lexer.rs
@@ -34,7 +34,10 @@ fn lex_float() {
 
 #[test]
 fn lex_string() {
-    let inputs = vec!["\"Blabla\"", "\"123\""];
+    let inputs = vec!["\"Blabla\"", "\"123\"", 
+                      "\'123\'", "\'This is still a string\'",
+                      "`this is backquted`", "`and this too`"
+                ];
     for i in inputs {
         assert_lex_rule!(Rule::string, i);
     }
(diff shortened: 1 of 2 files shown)


src/parser/tests/lexer.rs
```diff
@@ -34,7 +34,10 @@ fn lex_float() {
 
 #[test]
 fn lex_string() {
-    let inputs = vec!["\"Blabla\"", "\"123\""];
+    let inputs = vec!["\"Blabla\"", "\"123\"", 
+                      "\'123\'", "\'This is still a string\'",
+                      "`this is backquted`", "`and this too`"
+                ];
```
Keats: can you use block formatting for that? Like in the `lex_dotted_ident`

Keats:
> I tried not to deviate from the description of the issue, but I'm not sure if it intended for Rule::string to "decay" to the Rule::text when there is an unmatched quote(i.e "\"string\"" is a Rule::string, but "\"string"(note the missing quote) is a Rule::text.

Hmm that sounds like a bug to me, if you have time I'd appreciate a fix!

Keats:
@ragne are you still working on that?

Issue #233, opened 2017-11-22
Sort filter (and possibly some others)


stephanbuys:
Hi, would it be possible to implement the following filters:

1) `sort`, as per Jinja, the ability to either sort an array of strings or an array of objects according to a field.
2) `dedup` or `uniq`, ability to eliminate duplicates based on (1), or any sorted list.

Keats:
I want to add that but it's not super clear how that would work, my thoughts were:

- `{{ my_array | sort }}` will sort primitives only, if you have objects/tuple instead and use `sort`, it will do nothing I guess?
- `{{ my array | sort(attribute="age")` will sort an array of hashmap/objects by the given attribute

but I'm not 100% convinced yet

stephanbuys:
I actually really like your proposal, what bugs you?

Keats:
It looks better after seeing it written but tuples will not be sortable, which is probably ok I think

stephanbuys:
Well you could make the tuple index numerical, zero-indexed? Almost exactly like rust?
so `{{ my array | sort(attribute=0) }}` would be first element of dict or tuple - that would be very cool :-)

Keats:
I don't think you can just do that, sorting a tuple implies accessing it like `k.1` and you can't put a variable for the index. I think it's fair enough to just ignore tuples for now and focus on string/numbers/objects.

If anyone is willing to add it, it should be in the `0.11` branch in the `src/builtins/filters/array.rs` and the tests should be in the same file.

hoggetaylor:
I'll see if I can get something going for this.

Keats:
Shipped in 0.11, thanks @hoggetaylor !

PR #238, opened 2017-12-01
Add support for sort filter #233

hoggetaylor:
I'm submitting this as is to get some feedback.

Some things I'm currently unhappy with:
- An attribute that does not map to an actual field on the object fails silently.
- Sorting a list of numbers that contains NaN or infinity will cause a panic. (See all of the unwraps on line 72 of array.rs)

### DIFF
--- a/docs/content/docs/templates.md
+++ b/docs/content/docs/templates.md
@@ -482,6 +482,43 @@ Returns the length of an array or a string, 0 if the value is not an array.
 ### reverse
 Returns a reversed string or array.
 
+### sort
+Sorts an array into ascending order.
+
+The values in the array must be a sortable type:
+- numbers are sorted by their numerical value.
+- strings are sorted in alphabetical order.
+- arrays are sorted by their length.
+- bools are sorted as if false=0 and true=1
+
+If you need to sort a list of structs or tuples, use the `attribute`
+argument to specify which field to sort by.
+
+Example:
+
+Given `people` is an array of Person
+
+```rust
+struct Name(String, String);
+
+struct Person {
+    name: Name,
+    age: u32
+}
+```
+
+The `attribute` argument can be used to sort by last name:
+
+```jinja2
+{{ people | sort(attribute="name.1") }}
+```
+
+or by age:
+
+```jinja2
+{{ people | sort(attribute="age") }}
+```
+
 ### urlencode
 Percent-encodes a string.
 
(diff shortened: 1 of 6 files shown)


src/builtins/filters/array.rs
```diff
@@ -40,6 +41,53 @@ pub fn join(value: Value, args: HashMap<String, Value>) -> Result<Value> {
     Ok(to_value(&rendered.join(&sep))?)
 }
 
+/// Sorts the array in ascending order.
+/// Use the 'attribute' argument to define a field to sort by.
+/// Set the 'reverse' argument to sort in descending order.
```
Keats: There is already a `reverse` filter so there is no need for a `reverse` option

src/builtins/filters/array.rs
```diff
@@ -40,6 +41,53 @@ pub fn join(value: Value, args: HashMap<String, Value>) -> Result<Value> {
...
+    use Value::*;
+    use self::Ordering::*;
+    match (a, b) {
+        (&Null, &Null) => Equal,
+        (&Bool(ref a), &Bool(ref b)) => a.cmp(b),
+        (&Number(ref a), &Number(ref b)) => a.as_f64().unwrap().partial_cmp(&b.as_f64().unwrap()).unwrap(),
+        (&String(ref a), &String(ref b)) => a.cmp(b),
+        (&Array(ref a), &Array(ref b)) => a.len().cmp(&b.len()),
+        (&Object(ref a), &Object(ref b)) => a.len().cmp(&b.len()),
+        (a, b) => type_of(a).cmp(&type_of(b))
```
Keats: I don't think we should sort things that cannot be objectively sorted.

Keats:
cc @stephanbuys

Keats:
@hoggetaylor are you still working on that?

hoggetaylor:
Sorry I have been busy finishing up exams. I am still working in this. I hope to have something pushed up soon.

hoggetaylor:
commit 6e29de6 resolves all panics in the previous implementation. Also giving an invalid attribute will result in an Err. 

The code clones a bit more than I would like but I was having some lifetime issues that I'm not quite experienced enough to resolve.

Keats:
The sort looks good as well. I would maybe just put all the code unrelated to the sort filter in another file though.

@stephanbuys can you confirm that it is what you had in mind?

Keats:
Thanks! I think it's just missing the docs in `docs/content/docs/templates.md` and it's good to merge

docs/content/docs/templates.md
```diff
@@ -482,6 +482,43 @@ Returns the length of an array or a string, 0 if the value is not an array.
...
 Returns a reversed string or array.
 
+### sort
+Sorts an array into ascending order.
+
+The values in the array must be a sortable type:
+- numbers are sorted by their numerical value.
+- strings are sorted in alphabetical order.
+- arrays are sorted by their length.
+- bools are sorted as if false=0 and true=1
```
Keats: now that i think of it, does it ever makes sense to sort an array of bool?

hoggetaylor: It's probably a pretty rare use case. But I don't think it would hurt to have the ability to do it.

Keats:
Thanks a lot!

PR #241, opened 2017-12-23
regression test for math on a number output by the length filter

elahn:
I don't know tera well enough to find the cause of this regression from 0.10.10 to 0.11.beta, but here's a test case that reproduces it.

### DIFF
--- a/tests/render-success/math_length.html
+++ b/tests/render-success/math_length.html
@@ -1,0 +1,1 @@
+{{ reviews|length - 1 }}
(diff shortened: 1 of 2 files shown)


Keats:
Oops missed that
That's sadly the main breaking change from 0.10: math operation are not allowed anymore after filters. I still need to have a better look if i can make it work though because it's bugging me

Keats:
Added #244 if someone finds the time

Keats:
@elahn Can you try the v0.11 branch to see if everything works for you now?

Keats:
That's not a regression anymore starting from beta3 so closing it, thanks!

PR #252, opened 2018-01-28
Fix pluralize filter

LFalch:
Nouns take plural when the number is not one. This is true for a lot of languages (including English) although some languages like French take singular with zero. This isn't accounted for in this PR though right now.

This PR makes pluralize return the suffix if the number is 1 or -1.

### DIFF
--- a/src/builtins/filters/number.rs
+++ b/src/builtins/filters/number.rs
@@ -7,7 +7,7 @@ use humansize::{FileSize, file_size_opts};
 use errors::Result;
 
 
-/// Returns a suffix if the value is not equal to 1. Suffix defaults to `s`
+/// Returns a suffix if the value is not equal to ±1. Suffix defaults to `s`
 pub fn pluralize(value: Value, args: HashMap<String, Value>) -> Result<Value> {
     let num = try_get_value!("pluralize", "value", f64, value);
     let suffix = match args.get("suffix") {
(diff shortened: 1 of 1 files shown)


Keats:
Can you add a testcase that was previously failing? Other than that it looks good

src/builtins/filters/number.rs
```diff
@@ -7,16 +7,16 @@ use humansize::{FileSize, file_size_opts};
...
     let num = try_get_value!("pluralize", "value", f64, value);
     let suffix = match args.get("suffix") {
         Some(val) => try_get_value!("pluralize", "suffix", String, val),
         None => "s".to_string(),
     };
 
-    // English uses plural with zero or more than one thing
-    if num == 0.0 || num >= 2.0 {
+    // English uses plural when it isn't one
+    if num.abs() != 1. {
```
Keats: i'm not a native english speaker but is the `abs` really needed there? Is there even a rule for singular/plural for negative values?

LFalch: The rule is basically if it’s “one” before, it’s singular all other numerals (including decimal numbers) take plural. And in the case of -1, you still say “minus one”, so it takes singular. So it could also be `n == 1. || n == -1.`

Keats:
Thanks!

Issue #254, opened 2018-02-04
Addition of integers results in floats?


Lirusaito:
I'd wanted to create a filter that would take a slice of an array, given two usizes, so I went and made this function, and everything works out well when they're both literals. However, for my use case, I want to add two integer variables together (or an integer variable and an integer literal, the bug happens in both cases).

```jinja
{% set start = 0 %}
{% set end = start + 5 %}
{% for item in arr | slice(start=start, end=end) %}
{{item}}
{% endfor %}
```
Here's my function (I'm using rocket and was having a hard time with macro(I honestly don't remember why) so I just manually wrote it all):
```rust
fn slice(value: Value, args: HashMap<String, Value>) -> Result<Value, rocket_contrib::tera::Error> {
    let arr: Vec<Value> = match from_value::<Vec<Value>>(value.clone()) {
        Ok(val) => val,
        Err(_) => return Err(format!("Unable to get value vector for `slice`").into())
    };
    if arr.is_empty() { return Ok(arr.into()); }

    let start = match args.get("start") {
        Some(start) => match from_value::<usize>(start.to_owned()) {
            Ok(start) => start,
            Err(_) => return Err(format!("Argument `start` was not a `usize`: {}", start).into())
        },
        _ => return Err(format!("Unable to get `start` for `slice`").into())
    };
    if start > arr.len() { return Ok(Vec::<Value>::new().into()); } // If we have nothing to give, return empty.

    let mut end = match args.get("end") {
        Some(end) => match from_value::<usize>(end.to_owned()) {
            Ok(end) => end,
            Err(_) => return Err(format!("Argument `end` was not a `usize`: {}", end).into())
        },
        _ => return Err(format!("Unable to get `end` for `slice`").into())
    };
    if end > arr.len() { end = arr.len() }

    Ok(arr[start..end].into())
}
```
And this, of course, yields the error:
```rust
    => Error: Argument `end` was not a `usize`: 5.0
```

This error does not happen when I do
```jinja
{% set start = 0 %}
{% set end = 5 %}
{% for item in arr | slice(start=start, end=end) %}
{{item}}
{% endfor %}
```
or just
```jinja
{% for item in arr | slice(start=0, end=5) %}
{{item}}
{% endfor %}
```

Keats:
[...]

PR #257, opened 2018-02-06
0.11.3

Keats:


src/builtins/filters/array.rs
```diff
@@ -66,6 +66,29 @@ pub fn sort(value: Value, args: HashMap<String, Value>) -> Result<Value> {
...
+/// Use the `start` argument to define where to start (inclusive, default to `0`)
+/// and `end` argument to define where to stop (exclusive, default to the length of the array)
+/// `start` and `end` are 0-indexed
+pub fn slice(value: Value, args: HashMap<String, Value>) -> Result<Value> {
+    let mut arr = try_get_value!("slice", "value", Vec<Value>, value);
+    if arr.is_empty() {
+        return Ok(arr.into());
+    }
+
+    let start = try_get_value!("slice", "start", f64, args.get("start").unwrap_or(&0.0.into())) as usize;
```
bootandy: Forgive my curiosity, but why f64 and not an integer?

Keats: See https://github.com/Keats/tera/issues/254#issuecomment-363062924

Keats:
@lirusaito can you check if that works for your templates or if you are still running into macro issues?

Lirusaito:
Macro issues disappeared, I think I was chasing ghosts (if you mean rust macro issues), as for my personal macro issues, I just have the work arounds in place to deal with them.
I'll try this change tomorrow, unless I start my laptop up today, then probably today.

Lirusaito:
@Keats I went to test it with my personal templates, but they are .html.tera files set up for rocket, is there a quick bit of code I could use to make tera consume everything as though it didn't have any extensions? Ideally, I'd like to avoid mucking up too many files just to test, since I have other uncommitted work adjacent to the areas I'll be testing in.

Keats:
No easy way to do that no. You can cp the templates folder and do a regex
search/replace for extends/include/import to add the extension in the call

Lirusaito:
Oki, I think I'll cp then remove extensions.

Keats:
@Lirusaito ping

netlify.toml
```diff
@@ -0,0 +1,13 @@
+[build]
+    base    = "docs"
+    publish = "docs/public"
+    command = "gutenberg build"
+
+[context.production.environment]
```
calavera: if you don't want to use different binary versions in production and deploy previews, you can replace both `context.production.environment` and `context.deploy-preview.environment` with this single directive:

```
[build.environment]
  GUTENBERG_VERSION = "0.3.0"
```
that will apply to all contexts. If you need to use a specific version for a context, then you can override this value in the specific context.

Lirusaito:
My apologies for the delay, I'm not quite sure what you want me to be testing, so here's a list of all the issues:
+ 256 - Flow control with nested imports: Still an issue (I see that this is marked for next major, so that makes sense)
+ 255 - Macro files with imports: No longer an issue
+ Just for the heck of it, I also tried importing inside a macro itself, and it was still an unexpected tag, which makes sense.

Keats:
> Just for the heck of it, I also tried importing inside a macro itself, and it was still an unexpected tag, which makes sense.

I don't believe this has ever been supported.
Anyway, seems that I can do a release now, thanks!

PR #260, opened 2018-02-16
Allow variable access via square brackets

### LINKED ISSUE #173: Allow variable indexing in templates
Such as `my_variables[0]` (equivalent to my_variables.0) or `my_variables["hello"]`  (equivalent to my_variables.hello). The advantage being the indexed version can contain fields that are not valid Tera identifiers, such as `posts/tutorials` for example.

Potential issues (some might be non-existent after trying it):

- chained indexing: `myvar["hey"][0]`
- variable in indexes: `my_var[config.home]`. Could be not allowed and recommend to use `my_var | get(key=config.home)` instead

That might be a breaking change and should only be done once Pest reaches 1.0

bootandy:
Fixes #173

Allows tera to compute variables inside square brackets. 
Allows us to do things like: `var[i]`

If you have any suggestions / feedback please comment


src/parser/mod.rs
```diff
@@ -745,6 +745,8 @@ pub fn parse(input: &str) -> TeraResult<Vec<Node>> {
                     Rule::boolean => "`true` or `false`".to_string(),
                     Rule::ident => "an identifier".to_string(),
                     Rule::dotted_ident => "a dotted identifier (identifiers separated by `.`)".to_string(),
+                    Rule::dotted_square_bracket_ident => "an identifier (identifiers separated by `.` or `[]`s)".to_string(),
+                    Rule::square_brackets => "an identifier, string or integer inside `[]`s".to_string(),
```
Keats: looks like there is some overlap on the error messages which can cause some confusing results like a test below.

Maybe modify the `Rule::ident` message to be more precise and mention it can only be a-z etc?

src/parser/tests/errors.rs
```diff
@@ -53,7 +55,7 @@ fn wrong_end_block() {
         "{{ hey %}",
         &[
             "1:9",
-            "expected an integer, a float, a string, `true` or `false`, an identifier, a dotted identifier (identifiers separated by `.`), or an expression"
+            "expected an integer, a float, a string, `true` or `false`, an identifier, a dotted identifier (identifiers separated by `.`), an identifier (identifiers separated by `.` or `[]`s), or an expression"
```
Keats: the confusing message I was referring above: `an identifier, ... , an identifier (...),`

src/renderer/mod.rs
```diff
@@ -83,9 +83,51 @@ impl<'a> Renderer<'a> {
...
+
+            for sub_var in sub_vars_to_calc.iter() {
+                // Translate from variable name to variable value
+                let post_var = find_variable(context, sub_var.as_ref(), tpl_name).unwrap();
+                let post_var_as_str = {
+                    if post_var.is_string() {
+                        post_var.as_str().unwrap().to_string()
+                    } else if post_var.is_boolean() {
+                        post_var.as_bool().unwrap().to_string()
+                    } else if post_var.is_i64() {
```
Keats: add `f64` as well maybe? since that's the result of math operations in Tera

src/tera.rs
```diff
@@ -704,6 +705,49 @@ mod tests {
...
+        context.add("deep_map", &deep_map);
+        context.add("a", &vec!["true", "false"]);
+        context.add("zero", &0);
+
+        let result = Tera::one_off(
+            "{{deep_map['inner_map'][a[zero]]}}", &context, true
+        ).unwrap();
+        assert_eq!(result, "yes");
+    }
+
```
Keats: can you move all those tests in `renderer/tests/square_brackets_ident.rs` or something similar?

Keats:
It does seem to slow down the benches a bit though, it's probably worth looking for a`[` in the key name before doing all the work

src/renderer/mod.rs
bootandy: Yep. Fixing.

src/parser/tests/errors.rs
bootandy: I agree that this is a confusing message. But I'm not sure how best to fix it. As an error it is listing everything that it could expect to see unfortunately this means it lists identifier 3 times. 

I'll try to resolve this with the next commit - please let me know your thoughts on it

src/tera.rs
bootandy: will do

src/renderer/mod.rs
```diff
@@ -686,3 +728,54 @@ impl<'a> Renderer<'a> {
...
+            },
+            None => break,
+            _ => (),
+        }
+    }
+    results
+}
+
+#[test]
+fn test_pull_out_square_bracket() {
```
bootandy: @Keats should this test be moved with the ones below as well? - I'm conscious there were no other tests in this file

src/tera.rs
bootandy: Moved

bootandy:
Yes it does seem to have affected benches - I'll have a look at that too.

src/renderer/mod.rs
```diff
@@ -81,11 +81,62 @@ impl<'a> Renderer<'a> {
...
+                .replace("]", "")
+        }
+
         #[inline]
         fn find_variable(context: &Value, key: &str, tpl_name: &str) -> Result<Value> {
-            match context.pointer(&get_json_pointer(key)) {
+            let key_s = match key.contains('[') {
+                false => key.into(),
+                true => evaluate_sub_variable(context, key, tpl_name),
+            };
```
Keats: clippy will probably tell you to use `if else` rather than matching on true/false, a good idea here imo

src/renderer/mod.rs
```diff
@@ -686,3 +737,54 @@ impl<'a> Renderer<'a> {
...
+                                results.push(sub);
+                            }
+                            break;
+                        }
+                        inside_bracket.push(c.unwrap());
+                    }
+                }
+            },
+            None => break,
+            _ => (),
```
Keats: I don't believe this is reachable?

src/renderer/mod.rs
Keats: I think the tests and the function itself can be moved to another file in the module (`square_brackets.rs` or something similar). This test is only testing the function directly so it's ok to leave it next to the function

src/tera.rs
```diff
@@ -622,6 +622,7 @@ mod tests {
     use super::{Tera};
     use context::Context;
     use serde_json::{Map as JsonObject, Value as JsonValue};
+    use std::collections::HashMap;
```
Keats: is that needed?

src/renderer/mod.rs
bootandy: moving to new file

src/tera.rs
bootandy: good spot. I'll remove it

src/renderer/mod.rs
bootandy: I actually prefer this style. Why is if/else better?

But clippy told me to change it. And we should obey clippy so I'll change it.

src/renderer/mod.rs
bootandy: This is needed for the case of: `Some(_)`

bootandy:
I am suspicious of that test failure. It builds fine locally for me. Can we easily rerun the old green builds or clean the test environment?

Keats:
Tests still fail on both appveyor and travis but pass locally for me, weird. I'll have a look this afternoon

Keats:
Got it, run `cargo update`: a new version of pest has been released and has apparently changed some of the error results

bootandy:
Yep looks like you are right. That should be fixed on a different branch and merged in - as it breaks master now.

Keats:
Done. I'll play with it this weekend a bit to see how it feels 
Thanks a lot for the work!

Keats:
Quick feedback: if a variable is not found when using square brackets (`var[10]`), it will report the error as `var.10` instead of `var[10]`. This is easily fixed by using `key` instead of `key_s` in the error message.

If we have `var[b]` and `b` is not defined, the message will show `"Variable `b` not found in context ` but doesn't mention `var[b]`. Something like `"Variable `b` in var[b] not found in context ` would be easier to understand as an error. Not sure why

src/renderer/mod.rs
```diff
@@ -81,11 +83,63 @@ impl<'a> Renderer<'a> {
...
         }
 
+        #[inline]
+        fn evaluate_sub_variable(context: &Value, key: &str, tpl_name: &str) -> String {
+            let sub_vars_to_calc = pull_out_square_bracket(key);
+            let mut new_key = key.to_string();
+
+            for sub_var in sub_vars_to_calc.iter() {
+                // Translate from variable name to variable value
+                let post_var = find_variable(context, sub_var.as_ref(), tpl_name).unwrap();
```
Keats: why is this `unwrap`? I believe this can fail is the value is not present no?

bootandy:
Seems reasonable to me. I'll fix it up soon and update the pull request.

src/renderer/mod.rs
bootandy: fixed in a refactor

Keats:
Any chance you can remove my 2 commits from master from the history? Thanks for the work, I'll merge it when I release the next version in a couple of days or so

bootandy:
I pull --rebased from upstream master locally and then force pushed this branch - I assume this is what you meant yes?

Keats:
Yep that's perfect thanks

PR #271, opened 2018-03-16
Formatting

Keats:


Keats:
You can compare with the https://github.com/Keats/tera/tree/fmt2 branch for actual rustfmt changes on top of that one. I left out what I consider to be poor/useless formatting.

PR #272, opened 2018-03-20
Add loop controls (`break` and `continue`)

### LINKED ISSUE #267: Add loop controls
`break` and `continue`: see http://jinja.pocoo.org/docs/dev/templates/#loop-controls in Jinja2

This would avoid ugly things like https://github.com/Keats/book/blob/master/templates/page.html#L29-L43

aluminous:
Resolves #267.

Most rendering functions return `Result<String>`. I wanted to avoid reusing `Result` to indicate that a `break` or `continue` was encountered, so I changed the return type of some functions to `Result<LoopControl>`.

`LoopControl` is similar to `Result`, but doesn't indicate failure. This ensures that a break isn't ignored or permitted outside loops. Outside a loop where rendering must finish normally, `LoopControl.into_normal_result()` allows obtaining a `Result<String>` while checking break/continue was not encountered.

If you have a different approach in mind, or have other feedback, please let me know!

### DIFF
--- a/src/parser/tera.pest
+++ b/src/parser/tera.pest
@@ -169,17 +169,23 @@ endraw_tag = { tag_start ~ "endraw" ~ tag_end }
 raw_text   = { (!endraw_tag ~ any)* }
 raw        = !{ raw_tag ~ raw_text ~ endraw_tag }
 
-macro_filter_section = !{ filter_tag ~ macro_content* ~ endfilter_tag }
-block_filter_section = !{ filter_tag ~ block_content* ~ endfilter_tag }
-filter_section       = !{ filter_tag ~ content* ~ endfilter_tag }
+macro_filter_section = !{ filter_tag ~ macro_for_content* ~ endfilter_tag }
+block_filter_section = !{ filter_tag ~ block_for_content* ~ endfilter_tag }
+filter_section       = !{ filter_tag ~ for_content* ~ endfilter_tag }
+macro_for_filter_section = !{ filter_tag ~ macro_for_content* ~ endfilter_tag }
+block_for_filter_section = !{ filter_tag ~ block_for_content* ~ endfilter_tag }
+for_filter_section       = !{ filter_tag ~ for_content* ~ endfilter_tag }
 
-macro_forloop = ${ for_tag ~ macro_content* ~ endfor_tag }
-block_forloop = ${ for_tag ~ content* ~ endfor_tag }
-forloop       = ${ for_tag ~ block_content* ~ endfor_tag }
+macro_forloop = ${ for_tag ~ macro_for_content* ~ endfor_tag }
+block_forloop = ${ for_tag ~ for_content* ~ endfor_tag }
+forloop       = ${ for_tag ~ block_for_content* ~ endfor_tag }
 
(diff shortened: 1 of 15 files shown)


src/renderer/mod.rs
```diff
@@ -802,10 +826,42 @@ impl<'a> Renderer<'a> {
...
+    Continue(String),
+    Break(String),
+}
+
+impl LoopControl {
+    fn into_normal_result(self) -> Result<String> {
+        match self {
+            LoopControl::Normal(inner) => Ok(inner),
+            LoopControl::Break(_) => bail!("`break` appeared outside a loop"),
+            LoopControl::Continue(_) => bail!("`continue` appeared outside a loop"),
```
Keats: putting it in the grammar will avoid having to deal with errors at this point

aluminous:
Thanks for the review. I left a comment regarding the grammar.

I will look into adding state to ForLoop instead of using LoopControl.

aluminous:
I updated the code per your comments -- removing `LoopControl` seems like an improvement. It seems like multiple new types are required in the grammar, due to cases like the following:

```jinja
{% block foo %}
    block_content
    {% for i in list %}
        block_for_content
        {% if i.example %}
            If parsed as the existing `block_if`, this would be `block_content` -- the
            fact that this is a loop would be forgotten.
            {% break %}
        {% endif %}
    {% endfor %}
{% endblock foo %}
```

What are your thoughts on this? Maybe I'm over complicating things?

src/renderer/mod.rs
```diff
@@ -745,6 +771,13 @@ impl<'a> Renderer<'a> {
 
         for n in body {
             output.push_str(&self.render_node(n)?);
+            if let Some(for_loop) = self.current_for_loop() {
+                match for_loop.state {
+                    ForLoopState::Continue
+                    | ForLoopState::Break => break,
```
Keats: why is it breaking on continue?

src/parser/tera.pest
```diff
@@ -229,6 +271,25 @@ content = @{
...
+    block |
+    content_for_if |
+    forloop |
+    break_tag |
+    continue_tag |
+    for_filter_section |
+    raw |
+    text
+}
+
```
Keats: Yeah definitely too many changes. Only a few things are actually valid in a for loop: ifs, set/set_global, include, break, continue, raw, text, filter_section. Block, import macro,macro definition cannot be in it.

I think the current state might even be wrong as it allows some stuff that shouldn't be allowed. I'll fix it myself after merging, I'm not even entirely sure what's allowed/unallowed

src/renderer/mod.rs
aluminous: It breaks on continue to stop processing the remaining statements in the iteration. The next iteration of the for loop in the template still executes

src/parser/tera.pest
aluminous: Okay. I tested a block inside a loop on master and it didn't throw an error so I left things the way they were for now.

I'm still not sure how to avoid the other new for_content types (and for_ifs etc)

aluminous: Ah, I see.

I made another update, and after limiting the types allowed in for_content, the other new types could be simplified.

Keats:
I have absolutely no idea exactly how the grammar for forloops should look like so I'll merge that and spend some time on it when I can. Thanks!

PR #280, opened 2018-04-23
fix: don't truncate strings equal to length

jkcclemens:
Fix a panic from the unwrap below when the string length is equal to the truncate length.


Keats:
woops, thanks!

PR #285, opened 2018-05-16
User-defined escape function.

yakov-bakhmatov:
Tera is very nice template engine. But in my project I need custom escape function. So I added it to Tera, and I think it may be useful for other users. (Sorry for my English.)

### DIFF
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -675,22 +675,6 @@ mod tests {
     use context::Context;
     use serde_json::{Map as JsonObject, Value as JsonValue};
 
-    fn escape_c_string(input: &str) -> String {
-        let mut output = String::with_capacity(input.len() * 2);
-        for c in input.chars() {
-            match c {
-                '\'' => output.push_str("\\'"),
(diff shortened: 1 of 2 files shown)


src/tera.rs
```diff
@@ -638,6 +675,22 @@ mod tests {
     use context::Context;
     use serde_json::{Map as JsonObject, Value as JsonValue};
 
+    fn escape_c_string(input: &str) -> String {
```
Keats: when would this be used? I think this doesn't need to be in Tera directly

yakov-bakhmatov: `escape_c_string` is in `tera::tests` module, so it is not visible in production code.

yakov-bakhmatov: I moved `escape_c_string` into test function.

Keats: I missed that ;( sorry

Keats:
Looks good!
I'll merge it once I'm going to make a new release, shouldn't take too long

PR #288, opened 2018-05-16
Accept title case boolean literals ("True" and "False")

alex-berger:
As documented in http://jinja.pocoo.org/docs/2.10/templates/, Jinja2 supports lower case boolean literals as well as title case boolean literals.

In order to increase Tera's compatibility with Jinja2 respectively to reduce porting overhead when switching from Jinja2 to Tera this commit adds support for (python like) title case boolean literals to Tera.


Keats:
Thanks, that will be in the next version

PR #242, opened 2017-12-26
String concatenation support

lettenj61:
Thanks for warm advice and patience, I made parser / renderer work with concatenation, but only added 1 test. I think I should add more tests (especially for parser).

Issue #222, opened 2017-10-23
Support string concatenation as an operator


lettenj61:
Hello, I love Tera and Gutenberg. They're awesome!

I'm writing a Gutenberg theme, and I found that it will be useful if Tera supports string concatenation as we can in Jinja2, but I wonder is it acceptable or not.

I'm thinking of a template something like:

```jinja2
{{ "Hello" ~ " world" }}
```

I wrote some examples and unit tests on [my fork](https://github.com/Keats/tera/compare/master...lettenj61:string-concatenation), but I believe there's v0.11 coming so they may be outdated.

Keats:
The Jinja2 doc say the following:

```
+
Adds two objects together. Usually the objects are numbers, but if both are strings or lists, you can concatenate them this way. This, however, is not the preferred way to concatenate strings! For string concatenation, have a look-see at the ~ operator. {{ 1 + 1 }} is 2.
```

Any idea why `+` is not the preferred way?

> I wrote some examples and unit tests on my fork, but I believe there's v0.11 coming so they may be outdated.

Yeah, any work should be done on the 0.11 branch. It's mostly waiting for pest to release a stable version but it almost ready otherwise

lettenj61:
@Keats 

> Any idea why + is not the preferred way?

I feel totally OK with `+`, I just blindly followed Jinja2 syntax.

To be honest, I have no experience of Python / Django / Jinja2, so I have no idea why they discourage `+` for strings. I guess one of the reasons is that Python normal `operator` fails at runtime when you try to add non-numbers.

For Tera internals, if we use `+` for strings, we must treat `MathExpr`s specially at rendering when either of `lhs` or `rhs` would not be a number. My idea is based on introducing new `Concat` node for the AST, I'm not sure which way is the better.

I understood about next version. I'll work on `v0.11` branch.

Keats:
Hmm thinking about it we should maybe use `~` so we get the same syntax highlighting as Jinja2

Keats:
@lettenj61 do you need help with that?

lettenj61:
@Keats Sorry for the late, I'm working on it but I couldn't get much time. I hope I can open a PR in a next few days.

I agree with `~` for syntax highlighting purpose.

Keats:
Let me know if you have any questions with `pest` or `tera`!

verpeteren:
I like to use this feature in a project that I am working on.
(I want to concat  two variables into one macro argument)

[...]

PR #289, opened 2018-05-17
String Concatenation

travismiller:
Here's an updated pull request for adding concatenation support, continuing from @lettenj61's work. Review and feedback please!

## Related Issues and PRs

- #222
- #242


src/parser/tera.pest
```diff
@@ -78,7 +79,7 @@ dotted_square_bracket_ident = @{
 
 // boolean first so they are not caught as identifiers
 basic_val  = _{ boolean | test | macro_call | fn_call | dotted_square_bracket_ident | float | int | string }
-basic_op   = _{ op_plus | op_minus | op_times | op_slash | op_modulo }
+basic_op   = _{ op_plus | op_minus | op_times | op_slash | op_modulo | op_tilde }
```
Keats: This allows `~` in operations which is not great, ie `1 ~ 2` would be considered a concat but doesn't mean anything in the real world.

I view string concat more like `str_concat = { (string | dotted_square_bracket_ident) ~ "~" ~  string | dotted_square_bracket_ident)}` (maybe with `+` somewhere if we want multiple concat) and that `str_concat` rule being part of `basic_val`.

I haven't thought deeply about it so there might be some unforeseen issues with that though

src/parser/tests/parser.rs
```diff
@@ -127,6 +127,67 @@ fn parse_variable_tag_lit_math_expression() {
...
+#[test]
+fn parse_variable_tag_concat_strings() {
+    let ast = parse("{{ \"a\" ~ \"b\" }}").unwrap();
+
+    assert_eq!(
+        ast[0],
+        Node::VariableBlock(Expr::new(ExprVal::Concat(Concat {
+            lhs: Box::new(Expr::new(ExprVal::String("a".to_string()))),
+            rhs: Box::new(Expr::new(ExprVal::String("b".to_string()))),
+        }))),
```
Keats: in the case of 2 literal strings concat together, the parser should just create a String rather than a Concat since we can figure out the result at parsing time. The only time a Concat should be created is if one of the side is an ident since we can't resolve it at parsing time

src/parser/tera.pest
travismiller: I'm struggling with pest and I think I've found all of the dead ends with the suggested approach. If possible, I'd like to have a synchronous discussion on this.

Of note: the Jinja parser in Python appears to combine concatenation with the math operation parsing. https://github.com/pallets/jinja/blob/master/jinja2/parser.py#L494-L524

Keats: let me do a spike on my end first since i didn't really think things through in terms of code yet

Keats: It was easier than I thought, #293

PR #300, opened 2018-06-09
Added two benchmarks `big_loop_big_objects`, `macro_big_object`

patefacio:
I need to do more work to get all tests passing on a less-cloning rendering.
This is before/after for two new benchmarks.

------------------------

test bench_macro_big_object    ... bench:   1,794,697 ns/iter (+/- 135,373)
test bench_big_loop_big_object ... bench:   2,759,285 ns/iter (+/- 748,879)

------------------------
test bench_macro_big_object    ... bench:     663,081 ns/iter (+/- 68,959)
test bench_big_loop_big_object ... bench:     434,361 ns/iter (+/- 726,989)

### DIFF
--- a/benches/big_context.rs
+++ b/benches/big_context.rs
@@ -1,0 +1,96 @@
+#![feature(test)]
+extern crate test;
+extern crate tera;
+#[macro_use]
+extern crate serde_derive;
+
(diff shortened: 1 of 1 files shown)


Keats:
Thanks, will be helpful!

Issue #301, opened 2018-06-14
Range function and variable arguments


vdagonneau:
Hello ! 

If I'm not mistaken I may have stumbled upon a bug in the range function. At the very least it has a counterintuitive (to me) behavior.

### Minimal code that works

```
extern crate tera;

use tera::{Context, Tera};

fn main() {
    let mut context = Context::new();
    context.add("ctx_start", &0);
    context.add("ctx_end", &3);

    let result = Tera::one_off(
        "{% for i in range(start=ctx_start, end=ctx_end) %}{{ i }}\n{% endfor%}",
        &context,
        true
    );
    println!("{}", result.unwrap());
}
```

### Minimal code that doesn't work

```
extern crate tera;

use tera::{Context, Tera};

fn main() {
    let mut context = Context::new();
    context.add("ctx_start", &0);
    context.add("ctx_end", &3);

    let result = Tera::one_off(
        "{% set set_end = ctx_start + 3 %}{% for i in range(start=ctx_start, end=set_end) %}{{ i }}\n{% endfor%}",
        &context,
        true
    );
    println!("{}", result.unwrap());
}
```

The only difference is when I set a variable in the template itself. Am I missing something ?

Keats:
Note:

1. works normally
2. `set` returns a float and `end` only works with integers

The `range` function should try to parse it as int, that should be an easy fix if anyone is interested

vdagonneau:
I took a dive into the code thanks to your pointers.

I think the problem is that both `eval_expr_as_number` and `eval_as_number` are working only on `f64` internally. Probably because the JSON number can be either `f64`, `u64` or `i64` ?

I am working on a commit to work with Number instead and stop casting every number back to an `f64`. It seems to work for now but breaks on `NaN` (divide by zero).

Keats:
Thanks! If you do a PR, do it on the `next` branch

PR #302, opened 2018-06-15
Reworked the math evaluation to work on u64, i64 or f64

vdagonneau:
#301 This is a basic fix for the problem. It just solves it. It also now panics on NaN operations (divide by 0), I am not sure if it is the expected behavior. I also haven't provided tests yet.

### DIFF
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -241,10 +241,10 @@ impl<'a> Renderer<'a> {
 
     /// In some cases, we will have filters in lhs/rhs of a math expression
     /// `eval_as_number` only works on ExprVal rather than Expr
-    fn eval_expr_as_number(&mut self, expr: &Expr) -> Result<Number> {
+    fn eval_expr_as_number(&mut self, expr: &Expr) -> Result<Option<Number>> {
         if !expr.filters.is_empty() {
             match self.eval_expression(expr)? {
-                Value::Number(s) => Ok(s),
+                Value::Number(s) => Ok(Some(s)),
                 _ => bail!("Tried to do math with an expression not resulting in a number"),
             }
         } else {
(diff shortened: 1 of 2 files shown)


Keats:
Hmm I don't think panicking on NaN makes sense, we don't want any template rendering to panic. I think the current behaviour of returning NaN makes sense as it's not an error of the template engine directly - you could get a NaN purely from user data and it would make sense to display NaN rather than not being able to render the template

src/renderer/mod.rs
```diff
@@ -253,34 +253,106 @@ impl<'a> Renderer<'a> {
...
+                            let rr = r.as_i64().unwrap();
+                            Ok(Number::from(ll * rr))
+                        } else if l.is_u64() && r.is_u64() {
+                            let ll = l.as_u64().unwrap();
+                            let rr = r.as_u64().unwrap();
+                            Ok(Number::from(ll * rr))
+                        } else {
+                            let ll = l.as_f64().unwrap();
+                            let rr = r.as_f64().unwrap();
+			    Ok(Number::from_f64(ll * rr).unwrap())
```
Keats: wrong indent

src/renderer/mod.rs
```diff
@@ -253,34 +253,106 @@ impl<'a> Renderer<'a> {
...
+                    MathOperator::Div => {
+                        let ll = l.as_f64().unwrap();
+                        let rr = r.as_f64().unwrap();
+			let res = ll / rr;
+			if res.is_nan() {
+			    Err(From::from("NaN"))
+			} else {
+                            Ok(Number::from_f64(res).unwrap())
+			}
+                    },
```
Keats: wrong indent as well

src/renderer/mod.rs
```diff
@@ -253,34 +253,106 @@ impl<'a> Renderer<'a> {
...
+                    MathOperator::Add => { 
+                        if l.is_i64() && r.is_i64() {
+                            let ll = l.as_i64().unwrap();
+                            let rr = r.as_i64().unwrap();
+                            Ok(Number::from(ll + rr))
+                        } else if l.is_u64() && r.is_u64() {
+                            let ll = l.as_u64().unwrap();
+                            let rr = r.as_u64().unwrap();
+                            Ok(Number::from(ll + rr))
```
Keats: I believe in all cases other than `float` we can unwrap numbers as `i64` no?

vdagonneau: Well, I'm pretty sure you can't cast a u64 to i64.

For example, if the input number is std::i64::MAX + 1, it is a valid u64 but definitely not the same value as an i64.

The JSON Number type from serde stores either the f64, u64 or i64 based on the value provided. We need to get the correct one. I agree it is very verbose and I hope we can improve that, I'm open to suggestion.

For reference, the implementation of Number: https://github.com/serde-rs/json/blob/master/src/number.rs

vdagonneau:
I'll correct the indentation and I am also on my way to correct the two tests that are failing.

Keats:
Last thing I forgot to ask, can you add a test for the `range` issue you had?

vdagonneau:
It should be good to go now.

Keats:
Thanks!

Issue #264, opened 2018-02-28
Add examples with web servers


Keats:
That work on stable:

- hyper (0.12 is going to be released soon)
- actix

Keats:
There is one with actix now, the rest should be fairly similar so closing

PR #307, opened 2018-07-24
Examples - Actix Server

skswanke:
Partially solves #264 

Includes basic features such as logging, a 404 page and a nav. Enough to get someone going on a more complicated project. Also includes bootstrap (from cdn) and some of the classes in the html, this is easily removed however.

Compiles on Stable for both actix-web and rust (`1.27.1`).


Keats:
Thanks a lot!

PR #310, opened 2018-07-31
Add throw global function

Keats:


codesections:
I like this.  So it looks like the syntax for the original use-case I mentioned in keats/gutenberg#355 would become
```jinja2
        {% if page.summary %}
          {{ page.summary | safe }}
        {% else %}
          {% throw("Error: " ~ page.title ~ " did not include a summary") %}
        {% endif %}
```

Which would then display console message something like 

```
Building site...
-> Creating 29 pages (0 orphan), 3 sections, and processing 0 images
Error: Failed to render pager 2
Reason: Failed to render 'section.html': error while rendering a macro from the `blog` namespace
Reason: Error: Post1 did not include a summary
```
Is that about right?

Keats:
Close, that will be `{% throw(message="Error: " ~ page.title ~ " did not include a summary") %}`

Jieiku:
I made use of this feature to error check when required config.toml values are missing, for example:

```html
  {%- if config.compile_sass %}
  {%- set stylesheets=config.extra.stylesheets | default(value=[ "abridge.css" ]) -%}
  {%- if stylesheets %}{%- for i in stylesheets %}
  <link rel="stylesheet" href="{{     get_url(path=i                               , trailing_slash=false, cachebust=true) | safe }}" />
  {%- endfor %}{%- endif %}
  {%- else %}
    {{ throw(message="Error: Theme requires that compile_sass is set to true in config.toml") }}
  {%- endif %}
```

Jieiku:
actually I thought I figured this out, but it still does not work, will open a new issue report.

PR #317, opened 2018-08-14
Add as_str filter

Txuritan:
Sorry for the wait, I thought I might be able to finish this before my trip, guess not.

Anywho I just used the `ValueRender` trait to convert it into a string. I added it to the registry, made some tests (probably wasn't needed), and added it to the `Built-in filters` section in the docs.


Keats:
Looks good, thanks!

PR #321, opened 2018-08-22
Performance enhancement reducing clone

patefacio:


Cargo.toml
```diff
@@ -29,7 +29,10 @@ url = "1"
 humansize = "1"
 # used in date format filter
 chrono = "0.4"
+# used in instrument timing
```
Keats: the 3 dependencies below should be in dev-dependencies then?

src/bin/big_context_no_loop.rs
```diff
@@ -0,0 +1,104 @@
+extern crate tera;
```
Keats: is this file needed? can it be in bench instead?

src/renderer/ast_processor.rs
```diff
@@ -0,0 +1,1064 @@
+//! Responsible for walking ast and render implementation
+
+// --- module use statements ---
```
Keats: can you remove all those comments?

src/renderer/tests/errors.rs
```diff
@@ -109,12 +102,12 @@ fn error_unknown_index_variable() {
...
-    assert_eq!(
-        result.unwrap_err().iter().nth(1).unwrap().description(),
-        "Variable arr[a] can not be evaluated because: Variable `a` not found in context while rendering \'tpl\'"
-    );
+    let result_text = result_text(tera.render("tpl", &Context::new()));
+println!("ERROR: {}", result_text); // TODO
+    // assert_eq!(
+    //     result.unwrap_err().iter().nth(1).unwrap().description(),
+    //     "Variable arr[a] can not be evaluated because: Variable `a` not found in context while rendering \'tpl\'"
+    // );
```
Keats: it should be uncommented no?

Cargo.toml
```diff
@@ -29,7 +29,10 @@ url = "1"
...
+stopwatch = "^0.0.7"
+log = "^0.3.8"
```
Keats: was logging used only when debugging? can it be removed?

patefacio: Removed stopwatch dependency. I do use `log!` by creating a macro called `out!` that points to `log!` for dev but can switch to `println!` if desire is to see logs from testing, since testing does not support.

I can remove this logging if you'd like, but maybe best to leave until you have time to review, in case need to review code more generally and logging might help. Whichever you prefer.

src/renderer/tests/errors.rs
patefacio: Replaced this with a test.

src/renderer/ast_processor.rs
patefacio: Removed.

src/bin/big_context_no_loop.rs
patefacio: Removed file.

Cargo.toml
patefacio: I believe I effectively did a rebase. My first and my first pull request, so I may have screwed it up. I hope it helps you diff more easily - but I'm not seeing how it helps so maybe I missed a step.

src/renderer/context.rs
```diff
@@ -0,0 +1,150 @@
+use context::get_json_pointer;
+use serde_json::value::Value;
+
+// --- module struct definitions ---
```
Keats: sorry I meant to remove these comments in every files

patefacio: Ok, removed those.

Keats:
Since the diff was so big, I ended up rewriting this PR to make sure I understood it well enough to maintain the codebase properly: the result is in https://github.com/Keats/tera/pull/327

A few comments:
- `RefOrOwned` exists in the std as [Cow](https://doc.rust-lang.org/stable/std/borrow/enum.Cow.html)
- it uncovered a few bugs with macros that I fixed but still need some digging/additional tests 
- the new structure is WAY cleaner than the old one: much easier to understand where we are not to mention allocating way less

I need to check if there are other places I can avoid allocating but overall thanks a lot for that PR, I marked you as author of #327  as it is mostly your work after all.

patefacio:
This is great! Thanks. Someone recommended COW when I described the
use-case and when I read its docs it sounded very different than my needs.
So, it will be interesting to see how you used it.

PR #325, opened 2018-08-29
Rustfmt

Keats:


Keats:
Looks like with `use_small_heuristics = "max"` the result is very nice @bootandy

Issue #319, opened 2018-08-17
Do if block and default filter behave differently?


paulcmal:
While defining a macro like this:
```
{% macro banner(link="", small, medium="", large="", alt="") %}
{% if medium %}medium is set{% else %}medium is not set{% endif %}
{{ medium | default(value="medium is still not set (defaut)") }}
{% endmacro banner %}
```

The if block seems to evaluate to first and `medium is not set` is displayed. But `medium is still not set (default)` is not displayed, leading me to believe that the `default` filter has a different evaluation mechanism than if.

Practically, it means I can't use `default` as i intended to handle optional arguments, like so:
```
<source media="(max-width: 128em)" srcset="{{ utils::flexURL(path=large | default(value=medium | default(value=small))) }}">
```

Is there another mechanism to handle optional arguments in macros than setting them to a value that should evaluate to false? Should we change the current behavior? Or should i just write a few lines in the docs to make it more clear?

Keats:
They do behave a bit differently, `{{ some_unknown_value }}` would cause an error if the the variable is not set. The `default` filter is a short version of `{% if some_unknown_value %}{{ some_unknown_value }}{% else %}some default{% endif %}` for when a variable is not set at all, not when there a value is falsy. I don't think the behaviour needs to change.

A way to handle optional parameters would be to add the possibility to use `null` in parameters but in the case of this macro I would just define a `size` and use a `if` to handle the various cases

paulcmal:
> I guess a doc update would be the best path then

Thanks for the clarification. I think the new docs are still confusing as it says default is a shortcut for if. But like you said, this rule of thumb only works if the variable is not set in the context. I'll propose some modifications in a separate PR :)

> would be to add the possibility to use null in parameters but in the case of this macro I would just define a size and use a if to handle the various cases

The macro doesn't take a single size (sorry for not posting [the whole of it](https://paste.unixcorn.org/?bed235d4e46ce772#lAunHLwMvWoaE1bYEJFiZnIIbN75V7BFgtv7Vaqia2w=) before) but a variety of sources for different breakpoint (i.e. responsive images) where only the smallest is mandatory.

Adding a `null` value to be used in macro would help me out greatly in this case, as this was what i was trying to achieve with a default `""` value.

[...]

PR #330, opened 2018-09-06
make default filter behavior more explicit in docs

paulcmal:
Following #319 i'm proposing some changes to make the filter's behavior **even more** explicit :)


Keats:
Thanks!

Issue #332, opened 2018-09-07
Reorganizing the docs?


paulcmal:
Hey i love the project but i feel like the docs could be so much easier to read through. I think the Templates page especially is just really full of stuff. That could be divided into separate sections and some stuff (like detailed filters/tests/functions docs) could be moved to separate pages to reduce cognitive overhead when you know what info you're looking for.

I don't have an idea yet for the whole organization of the docs, but for the templates page I'd propose a page (or section) organized like this:
```
# Templates

## Introduction

### Tera basics (comments, values, operations)
### Raw
### Whitespace control

## Data structures

### Literals
### Variables (+ link to assignments)
### Expressions

## Manipulating data

### Assignments
### Filters + list of built-in (+ link to more docs)
### Tests + list of built-in (+ link to more docs)
### Global functions + list of built-in (+ link to more docs)
→ Need more, use macros (link to macro section)

## Control structures

### Conditions (if)
### Loops (for)

## Template interactions

### Include
### Inheritance

## Extending templates

### Macros
### Custom global functions
### Custom tests
### Custom filters
```

What do you think?

That would mean changing the way the sidebar menu currently works, among other things. I'd be happy to contribute this and try and do more contributions to the docs. Thank you so much for building this amazing templating engine and taking the time to write basic documentation! <3 I hope we can work on making it clearer and more welcoming to newcomers :)

Keats:
Thanks for the kind words!

I like having a single page, it makes CTRL+f easy. The original inspiration was http://jinja.pocoo.org/docs/2.10/templates/

The structure looks ok but a few things need to be changed imo:

- Inheritance should have its own ## level
- include is in control structure
- I kind of like having the built-ins at the end like in the Jinja2 doc so they don't clutter too much the middle of the page

paulcmal:
Ok i'll start working on that :)

> I like having a single page, it makes CTRL+f easy.

That's not always true. I love to ^f around too, but when you have so much information on a single page you can get a lot of false positives and you have to scroll around (that's less of a problem if the menu is well-organized).

> The original inspiration

Those docs look good indeed. The main difference is there navigation menu goes a level deeper, but i think that's a big difference in terms of usability (given the organization of the content). If we can arrange to reorganize the pages making third-level heading more relevant, would you accept a PR to the current navigation menu?

> I kind of like having the built-ins at the end like in the Jinja2 doc so they don't clutter too much the middle of the page

[...]

PR #335, opened 2018-09-10
Docs reorganize

paulcmal:
Hey, as discussed in #332 i started reorganizing the templates docs and added a third level for the navigation menu with toggle buttons. The usability could be improved, as currently expanding the submenu will scroll up/down to said submenu. I still consider it way clearer and better than it was before :)

I added exactly one sentence in the docs, in the Variables part:
```
If you'd like to define your own variables, see the [Assignments](#assignments) section.
````

One workaround for the scrolling issue could be to have all submenu anchors generated on the H1 level so that for opening any submenu inside for instance "Templates" the scrolling would stick on that level.

Any feedback? :)

### DIFF
--- a/docs/content/docs/templates.md
+++ b/docs/content/docs/templates.md
@@ -516,7 +516,7 @@ which also contains a `super()` so we render the `hey` block of the `grandparent
 
 The end result of that rendering (not counting whitespace) will be: "dad says hi and grandma says hello sincerely with love".
 
-## Built-in tools
+## Built-ins
 
 ### Built-in filters
 
--- a/docs/content/docs/templates.md
+++ b/docs/content/docs/templates.md
@@ -516,35 +516,33 @@ which also contains a `super()` so we render the `hey` block of the `grandparent
 
 The end result of that rendering (not counting whitespace) will be: "dad says hi and grandma says hello sincerely with love".
 
-## Built-in tools
-
-### Built-in filters
+## Built-in filters
 
 Tera has the following filters built-in:
 
-#### lower
+### lower
 Lowercase a string
 
-#### wordcount
+### wordcount
 Returns number of words in a string
 
-#### capitalize
+### capitalize
 Returns the string with all its character lowercased apart from the first char which is uppercased.
 
-#### replace
+### replace
 Takes 2 mandatory string named arguments: `from` and `to`. It will return a string with all instances of
 the `from` string with the `to` string.
 
 Example: `{{ name | replace(from="Robert", to="Bob")}}`
 
-#### addslashes
+### addslashes
 Adds slashes before quotes.
 
 Example: `{{ value | addslashes }}`
 
 If value is "I'm using Tera", the output will be "I\'m using Tera".
 
-#### slugify
+### slugify
 Transform a string into ASCII, lowercase it, trim it, converts spaces to hyphens and
 remove all characters that are not numbers, lowercase letters or hyphens.
 
@@ -552,17 +550,17 @@ Example: `{{ value | slugify }}`
 
 If value is "-Hello world! ", the output will be "hello-world".
 
-#### title
+### title
 Capitalizes each word inside a sentence.
 
 Example: `{{ value | title }}`
 
 If value is "foo  bar", the output will be "Foo  Bar".
 
-#### trim
+### trim
 Remove leading and trailing whitespace if the variable is a string.
 
-#### truncate
+### truncate
 Truncates a string to the indicated length. If the string has a smaller length than
 the `length` argument, the string is returned as is.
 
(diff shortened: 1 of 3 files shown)


Keats:
Few comments:

- we need to remove `| reverse` in `docs.html` when iterating on `section.pages (forgot to do that myself)
- I would have every menu open instead of toggling them on and off
- `default` filter has the wrong heading level
- `Conditions (If)` -> `if`
- `Loops (For)` -> `for`
- Extending templates and having macros in it is confusing, I would rather have macros in Control Structures and remove the Extending template menu

paulcmal:
The rest I'm 100% ok with but

> I would have every menu open instead of toggling them on and off

We can, but it's still a lot of information to browse (cognitively speaking). I think at some point it will really help to do something about it, whether breaking into more pages or this kind of third-level expansion. I mean, i understand that you like the (mostly) single-page docs. Wouldn't a `man` page would be appropriate for you? While letting us mere mortals use web-technologies to link pages together? :P

When i say "at some point" i mean the docs will only grow bigger. Even without features being added, there's already a lot to clarify and make more friendly to newcomers.

paulcmal:
OK I did it this way. For the current page, the menu is 3-levels deep and doesn't have controls. For other pages, it's 2-levels deep with controls for the third level. Quick notes:
- Menu controls are invisible to screen readers (with `aria-hidden` property) so they can have a nice menu without additional cruft
- Navigation menu should be wrapped in a <nav> for accessibility purposes
- Menu expansion could take place with an camouflaged checkbox so as not to mess with scrolling

If the rest is okay with you I'll take care of those last two points before you can merge :)

Keats:
Ah just noticed something: built-in {filters, tests, global functions} should be at the top level, there isn't really a good name to group them.

I still don't like the (+) approach though, it would work with JavaScript but doing it via CSS triggers anchor jumping and is not working properly on all pages. Also I would expect to be able to close by reclicking on the (+), which would have turned into a (-) probably. I don't believe we can get a satisfying UX without JS so I would remove it completely and leave the full menu open.
Another issue of not showing the full hierarchy is that I would need to click around to know where is what I am looking for: for example assignments  are under `Manipulating data` here but under `Control structures` in the Jinja2 docs.

paulcmal:
> is not working properly on all pages

What's not working properly? Could you be more specific?

> triggers anchor jumping (…) would have turned into a (-)

I addressed those two issues in my latest commits. Is this outcome okay by your usual JS standards? :D

> I would need to click around

Unless you're on the templates page already, in which case the menu will be expanded already. Is it such a concern? Having too many entries in a menu is a real accessibility/usability concern. I'm not sure hesitating between two sections is such a big deal in comparison?

> built-in {filters, tests, global functions} should be at the top level

I'll address this in a next commit and then in my opinion the PR would be ready for merging, unless you have more concerns :)

Keats:
After a small round of user testing:

- much better than before
- they expected clicking on the link (like Data structures) to show what's below because of the (+)
- to fix that, it would need a `cursor: pointer` and a hover state to make it different from the header.

From me:

- the footer needs a `position: fixed; bottom: 0;` otherwise it's floating around on pages with little content
- use square brackets instead (+) -> [+] imo
- now that builtins are at the top level, there is an entry for each filter/test/fn name which makes the menu waaaay too long. I would rather have some kind of hack to not show them in the menu.

> I think global functions implies there's non-global functions (other than macros) which is confusing. 

I agree, I'm considering ading a rename from global function to function for the v1 breaking changes.

paulcmal:
> they expected clicking on the link (like Data structures) to show what's below

Should it? Personally, I've always found it an anti-pattern in navigation menus to try and click a page and suddenly many children pop up and mess with the layout, especially on mobile.

> a hover state

The best i could come up with is `font-weight: bolder;`. I've experimented other things but i'm not a proper graphics designer so i'll stick with that and let someone improve upon :)

> the footer

Sorry what footer are you referring to? I didn't notice any problems on Firefox on desktop, although responsive view is a disaster (that's a quest for another day :D). Do you have a screenshot or something?

> I would rather have some kind of hack to not show them in the menu.

How about having a `Built-in tools` top-level as in my original proposal? That was its whole purpose in the first place :)

> I'm considering ading a rename from global function to function

I'll change the heading to have it fit properly, and leave the actual content as is for the moment, if that's okay with you.

Keats:
> Should it? Personally, I've always found it an anti-pattern in navigation menus to try and click a page and suddenly many children pop up and mess with the layout, especially on mobile.

I think it's because in many places the [+]/(+) is the way to expand and the click only does expansion, the title is not a link itself.

> Sorry what footer are you referring to?

I'll fix that and the responsiveness, don't worry about it.

> How about having a Built-in tools top-level as in my original proposal?

Maybe just call it Built-ins then

paulcmal:
> in many places

True. Well anyway if that's a behavior you'd like to achieve in a theme some day, you just have to wrap the link in the label and the whole of it gets clickable :)

> Maybe just call it Built-ins then

Done. Any more concerns to address? If not feel free the squash the commits :)

Keats:
Thanks!

paulcmal:
:heart:

PR #337, opened 2018-09-11
Fix set_global context in macros and for loops

### LINKED ISSUE #336: set_global broken in macros
Hi, I believe the `set_global` function doesn't work as it should in macros. It does define the variable in the parent's context, but it is not accessible from the macro context. The most simple way to reproduce is:
```
{% macro test_global() %}
    {% set_global output = "Gutenberg!" %}
    {{ output }}
{% endmacro test_global %}
```

Produces:
```
Error: Failed to render page '/home/user/Site/content/_common/footer.md'
Reason: Failed to render 'mytheme/templates/page.html': error while rendering macro `macros::test_global` (error happened in a parent template)
Reason: Variable `output` not found in context while rendering 'mytheme/templates/widgets/macros.html'
```

The value is accessible from `page.html` and `index.html` though (the former extends the latter).

PS: Here the `error happened in a parent template` is inaccurate. We're not exactly extending the template that failed, but rather importing it so the wording probably should be different :)

EDIT: I'm running gutenberg 0.4.2 and tera 0.11.5

paulcmal:
This should fix #336. At least according to my understanding of what a global context should be :D 
Please tell me whether my interpretation is correct. If so, I'll add a commit to make docs more clear about this before merging :)

### DIFF
--- a/src/renderer/call_stack.rs
+++ b/src/renderer/call_stack.rs
@@ -68,9 +68,21 @@ impl<'a> CallStack<'a> {
         self.stack.push(StackFrame::new_include(name, tpl));
     }
 
-    /// Returns mutable reference to current `StackFrame`
-    pub fn origin_frame_mut(self: &mut Self) -> &mut StackFrame<'a> {
-        self.stack.first_mut().expect("Origin frame missing")
+    /// Returns mutable reference to global `StackFrame`
+    /// i.e gets first stack outside current for loops
+    pub fn global_frame_mut(self: &mut Self) -> &mut StackFrame<'a> {
+        if self.current_frame().kind == FrameType::ForLoop {
+            for stack_frame in self.stack.iter_mut().rev(){
+                // walk up the parent stacks until we meet the current template
+                if stack_frame.kind != FrameType::ForLoop {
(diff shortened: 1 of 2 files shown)


paulcmal:
Quick notes:

- i took the liberty to use raw strings (with `#` delimiters) instead of using normal strings in the raw templates… because it was just easier for my brain to deal with those :D 
- there's a lot of different unit tests with the same boilerplate code; would you take a PR to turn that into a single parameterized test that takes a map of strings (test names for output in case of failure) to a vec structure as is currently? or is it preferable to keep all tests as they are?

Keats:
> there's a lot of different unit tests with the same boilerplate code; would you take a PR to turn that into a single parameterized test that takes a map of strings (test names for output in case of failure) to a vec structure as is currently? or is it preferable to keep all tests as they are?

I use parameterized test a lot when the inputs are small but in these cases I like having separate tests.


The PR looks good! One thing it reminded me is that the behaviour of `set/set_global` in `include` is currently not specified I believe but that's another problem.
I'll have a look at the docs PR so you don't get conflict with your other PR before adding the doc clarification.

Issue #297, opened 2018-05-25
Consider remove error_chain dependency


svartalf:
Related to https://github.com/rust-lang-nursery/error-chain/issues/240, `tera::Error` instances are `!Sync`. 

For example, it is impossible to return `tera::Error` directly from the `actix-web` handler, and it is easier just to suppress it: https://github.com/actix/examples/blob/master/template_tera/src/main.rs#L28

Since `tera::Error` is quite simple, it might be worth to replace `error_chain!` call with a plain enum or use the [failure](https://crates.io/crates/failure) crate.

Keats:
That's a breaking change so for the next major version.

Related recent discussion: https://www.reddit.com/r/rust/comments/8lt8k6/do_i_really_need_failureerrorchain/

Failure seems nice but not really actively maintained so maybe going the plain enum would be better, depending on the timeline for the next major version

Keats:
Looks like some of Failure is coming to Std so it would be better to wait before making a breaking change: https://www.reddit.com/r/rust/comments/92kozl/feedback_needed_upcoming_failure_012_release/e36pf5w/?context=0

Keats:
So the way to go is probably plain std Error enum since using `failure` would probably be a dead end as things get moved to std. I am not sure how to provide the same UX as with error-chain though where errors are chained pretty seamlessly to give the whole context.

Once we have it working, we can also add more error types like the ones from #322

PR #352, opened 2018-11-07
remove error-chain

andy128k:
Addresses #297

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,6 @@ serde = "1.0"
 serde_json = "1.0"
 pest = "2"
 pest_derive = "2"
-error-chain = "0.12"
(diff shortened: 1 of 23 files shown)


Keats:
Amazing, thanks!
The CI failure is due to the addition of `source` in `Error`  (:walking_man:) which is only stable since 1.30 so that raises the min version of Tera to 1.30, which is fine.

I will review the changes tomorrow or this weekend, thanks again!

Keats:
Thanks again!

Keats:
Playing with it locally it looks like we still need to impl Iterate on the Error somehow but otherwise it works fine!

Keats:
Added the iteration on the v1 branch so it should be fine!

Issue #299, opened 2018-06-04
Context for global functions


max-frai:
Hello, I use fluent-rs for internationalization.
For now tera supports to register only global static functions without context which help process incoming data, but what if I need some additional context data?
In template:
`{{ global_lang(key_name) }}`
In code:
`context.add("key_name", "localization-key-name")`

So I want to call `global_lang` with `localization-key-name` string as argument and inside access external object which stores all locale keys, etc and return result string data.

max-frai:
Seems like this is related #219

Keats:
I'd be happy to have a PR for #219, especially if it can fix your issues

andy128k:
`Arc` to the rescue. [This example](https://github.com/Keats/tera/blob/v1/src/renderer/tests/basic.rs#L672) shows how to deal with it. What do I miss here?

Keats:
`Next` doesn't contain a pointer. Here is an example of a Zola fn:

```rust
#[derive(Debug)]
pub struct GetPage<'a> {
    pages: Arc<HashMap<&'a String, SerializingPage<'a>>>,
}
impl<'a> GetPage<'a> {
    pub fn new(library: &'a Library) -> Self {
        let mut pages = HashMap::new();
        for page in library.pages_values() {
            pages.insert(
                &page.file.relative,
                library.get_page(&page.file.path).unwrap().to_serialized(library),
            );
        }

        Self {pages: Arc::new(pages)}
    }
}
impl<'a> Function for GetPage<'a> {
    fn call(&self, args: &HashMap<String, Value>) -> Result<Value> {
        let path = required_arg!(
            String,
            args.get("path"),
            "`get_page` requires a `path` argument with a string value"
        );
        match self.pages.get(&path) {
            Some(p) => Ok(to_value(p).unwrap()),
            None => Err(format!("Page `{}` not found.", path).into()),
        }
    }
}
```
Trying to register it gives a conflicting lifetime error because:

```rust
    pub fn register_tera_global_fns(&mut self) {
        self.tera.register_function("trans", global_fns::Trans::new(self.config.clone()));
        self.tera.register_function("get_page", global_fns::GetPage::new(&self.library));
}
```
The lifetime of `&self.library` will end and won't outlive the `'static`. I don't know if there is a way to fix that tbh

Keats:
If we can tie the lifetime to the traits to the Tera instance instead of `'static`, it might work?

andy128k:
[...]

PR #355, opened 2018-11-12
Issue 219 and 299

andy128k:


src/builtins/filters/mod.rs
```diff
@@ -10,4 +10,13 @@ pub mod object;
 pub mod string;
 
 /// The filter function type definition
-pub type FilterFn = fn(Value, HashMap<String, Value>) -> Result<Value>;
+pub trait FilterFn {
```
Keats: I don't really like having the `Fn` bit in the type anymore, do you think `Filter` would be ok? (and same for `Test`). It could maybe be confused with https://doc.rust-lang.org/std/iter/struct.Filter.html but that is a struct and it shouldn't be an issue since people can use `tera::Filter` or `import tera::Filter as TeraFilter` for disambiguation.

src/builtins/functions.rs
```diff
@@ -6,117 +6,120 @@ use serde_json::value::{from_value, to_value, Value};
...
-            },
-            None => 0,
-        };
-        let step_by = match args.get("step_by") {
-            Some(val) => match from_value::<usize>(val.clone()) {
-                Ok(v) => v,
-                Err(_) => {
-                    return Err(Error::msg(format!(
-                    "Global function `range` received step_by={} but `step` can only be a number",
+pub trait GlobalFn {
```
Keats: One of the other change is to completely remove the `global` part of global function from the API. since there are no local functions in Tera, it doesn't make too much sense.
Of course in this case we can't use `Fn` as it exists in the std...

Maybe `TeraFunction`, `TeraFilter` and `TeraTest`? Or `Test`, `Filter` and `Function` and recommend people to use them prefixed like `tera::Test` etc to avoid conflicts. I'm tempted by the second approach but not convinced at all

src/tera.rs
```diff
@@ -461,17 +462,17 @@ impl Tera {
     /// If a filter with that name already exists, it will be overwritten
     ///
     /// ```rust,ignore
-    /// tera.register_filter("upper", string::upper);
+    /// tera.register_filter("upper", &make_filter_fn(string::upper);
     /// ```
-    pub fn register_filter(&mut self, name: &str, filter: FilterFn) {
-        self.filters.insert(name.to_string(), filter);
+    pub fn register_filter(&mut self, name: &str, filter: &Rc<dyn FilterFn>) {
```
Keats: I haven't used `impl Trait` yet but is not possible to have `pub fn register_filter(&mut self, name: &str, filter: &impl FilterFn)` and do the `Rc` ourselves?

andy128k: I don't think so. To make an Rc register-functions should get ownership of a function.

andy128k: Also, Rc is a part of public API, because of public fields of struct `Tera`.

Keats: Ah yes I forgot it can now have a context and isn't just a function

Keats:
cc @SergioBenitez a bit late but you might be interested in that PR since you opened the original issue

src/builtins/functions.rs
Keats: Agreed

examples/basic/main.rs
```diff
@@ -1,17 +1,15 @@
...
 use std::collections::HashMap;
 
-use tera::{Tera, Context, Result};
+use tera::{Tera, Context, make_filter, Result};
 use serde_json::value::{Value, to_value};
 
 
-lazy_static! {
-    pub static ref TEMPLATES: Tera = {
+thread_local! {
```
Keats: Not being `lazy_static` anymore makes me sad :(
It complicates/slows down the use for 99% of Tera users, I am guessing using `Arc` is a bad idea?

I'll play with this branch on Gutenberg to see the impact

Keats: I mean since we take `'static`, shouldn't be with be able to use `Arc` instead of `Rc` to make it `Sync` @andy128k ?

andy128k: Done

Keats:
Thanks!
I'll try it with Zola and see how it goes

Keats:
@andy128k 
I ended up removing the `make_filter` etc fns in the v1-2 branch. The issue I can see if that things like https://github.com/Keats/tera/commit/05a150d357554a093299185d4e5fccfa94a2b249#diff-e4bce7021c59d29e4cdc5243a88805e5 don't work (same with the make_filter etc).

```bash
error[E0277]: expected a `std::ops::Fn<(&std::collections::HashMap<std::string::String, serde_json::Value>,)>` closure, found `renderer::tests::basic::NEXT_GLOBAL`
   --> src/renderer/tests/basic.rs:686:10
    |
686 |     tera.register_function("get_next_borrowed", &NEXT_GLOBAL);
    |          ^^^^^^^^^^^^^^^^^ expected an `Fn<(&std::collections::HashMap<std::string::String, serde_json::Value>,)>` closure, found `renderer::tests::basic::NEXT_GLOBAL`
    |
    = help: the trait `for<'r> std::ops::Fn<(&'r std::collections::HashMap<std::string::String, serde_json::Value>,)>` is not implemented for `renderer::tests::basic::NEXT_GLOBAL`
    = note: required because of the requirements on the impl of `std::ops::FnOnce<(&std::collections::HashMap<std::string::String, serde_json::Value>,)>` for `&renderer::tests::basic::NEXT_GLOBAL`
    = note: required because of the requirements on the impl of `builtins::functions::Function` for `&renderer::tests::basic::NEXT_GLOBAL`
```

This severely limits the use for stateful functions/filters/tests :/

andy128k:
@Keats Personally I didn't like those make_* functions. Interface is much better without them.
I don't see an issue with borrowed function as a problem. This example looks artificial.

`Arc` could also be removed. The only reason for it is a method `extend`.
`Tera` itself could be an `Arc` and instead of copying internals it could delegate calls to nested one.

Keats:
I don't think it's that artificial.
To take #299 as example, is it likely that you want to load all your translations in a `lazy_static` once and impl Function on it to be used in Tera. If this is not working, you will need clone the struct (which could be very expensive - lots of strings) just to pass it to Tera.

andy128k:
Yeah. but this example is still doable. Like
```rust
struct Translations(Arc<ExpensiveStructure>);

impl tera::Function for Translations { ... }

lazy_static! {
    static ref TRANSLATIONS: ExpensiveStructure = load_translations!();
    static ref TRANSLATIONS_REF: Arc<ExpensiveStructure> = Translations(Arc::new(TRANSLATIONS));
}
```

Keats:
The issue is that passing it to `register_function` doesn't work. The trait is implemented correctly since `tera.register_function("get_next", Next(AtomicUsize::new(1)));` is the line above and it works.
I'm missing something there

PR #356, opened 2018-11-12
fix result of an array filter for empty input

andy128k:


src/builtins/filters/array.rs
```diff
@@ -114,7 +114,7 @@ pub fn group_by(value: Value, args: HashMap<String, Value>) -> Result<Value> {
 pub fn filter(value: Value, args: HashMap<String, Value>) -> Result<Value> {
     let mut arr = try_get_value!("filter", "value", Vec<Value>, value);
     if arr.is_empty() {
-        return Ok(Map::new().into());
+        return Ok(arr.into());
```
Keats: Good catch!
We can even do `Value::Array(Vec::new())` if we want to avoid the conversion but not sure it makes sense

PR #357, opened 2018-11-12
fix object get filter

andy128k:


Keats:
That was a big one :o 
Thanks!

Issue #308, opened 2018-07-28
Support for "is not" in tests


anowlcalledjosh:
Jinja2 supports using `is not` for tests, like so:

```html+jinja
{% if 4 is not divisibleby(2) %}
everything is broken
{% endif %}
```

It would be awesome if Tera could support the same! `if not 4 is divisibleby(2)` just doesn't look as nice :)

Keats:
I'm still not sure about that, leaving it open to collect some feedback.

jcgruenhage:
> `is not` does look better but with this change `if not 4 is not divisibleby(2)` would also be valid and that does look ugly :/

Sure, it gives people more options to make ugly expressions, but it also gives them options to avoid them, so I'd say this would be a change for the better.

Issue #331, opened 2018-09-06
v1 release


Keats:
v1 should be released after the edition 2018 has been released to ensure everything works properly.
Beta versions can be released before that though in order to get all the changes a bit battle-tested, such as using it in Gutenberg.

Blockers:

- [x] #328 
- [x] #297 
- [x] #322 
- [x] #256 
- [x] #219 (#299 is related)
- [x] #212 (ensure globwalk works well on every OS)
- [x] remove `Context::add` (deprecated for now)
- [x] deprecate  `global_function` for `function`
- [x] remove `global_function` everywhere
- [x] remove `compile_templates!`
- [ ] look into not using serde_json
- [x] #374

Undecided:

- [ ] #320 this is a backward compatible change so no hurry
- [ ] #265 the usecase is pretty narrow, need to see if it the changes to the code to allow that are worth doing
- [x] #308 I would err into "no" because of the potentially weird/unreadable syntax it would allow
- [ ] #340 would be a huge work (I think) I don't have the time for right now

Keats:
If you're reading this and wondering whether you should use the alpha release on crates.io, the answer is no. Wait for the v1 release as there will breaking changes in between alpha releases

Keats:
Looks like it's mostly ready to go but there's still #410, #404, #375 and #368 I'd like to get in though.

None of them are huge time sink. The other issues can be added after 1.0.

Anything else anyone is missing and which would be a breaking change?

Keats:
Anything breaking missing for a v1 release? Otherwise it's going to happen next week.

Keats:
v1 is released!

PR #360, opened 2018-11-19
add support for 'is not' operator

andy128k:
Closes issue #308 

I know this feature is controversial but it is mentioned in #331 so I tried to implement it (just because i love parsers <3).
Feel free to reject this.


Keats:
Thanks, I'll give it a bit of thoughts but I think it will be part of v1

Keats:
Looks like this issue didn't like my force push, can you cherry pick the last commit and force push after rebase?

PR #361, opened 2018-11-24
Add `nth` filter

### LINKED ISSUE #358: Add nth filter
Looks like it is a glaring omission for array filters because we can index but in some cases that would avoid setting an intermediate variable:

```
{% set grandparent = get_section(path=page.ancestors | reverse | nth(index=1) %}
```
With a 0-based indexing

RealOrangeOne:
Fixes #358 

Whilst implementing this, I had a thought of an alternative implementation. Rather than `nth`, extending `get` to allow an `index` argument, but I'm not sure how implicit that gets, or how that affects the implementation


src/builtins/filters/array.rs
```diff
@@ -6,6 +6,24 @@ use errors::Result;
...
+/// If the array is empty, returns empty string
+pub fn nth(value: Value, args: HashMap<String, Value>) -> Result<Value> {
+    let arr = try_get_value!("nth", "value", Vec<Value>, value);
+
+    if arr.is_empty() {
+        return Ok(to_value("").unwrap());
+    }
+
+    let index = match args.get("index") {
+        Some(val) => try_get_value!("nth", "index", usize, val),
```
Keats: `index` is a bit long here, how about `n`?

src/builtins/filters/array.rs
```diff
@@ -6,6 +6,24 @@ use errors::Result;
 use serde_json::value::{to_value, Map, Value};
 use sort_utils::get_sort_strategy_for_type;
 
+
+/// Returns the nth value of an array
+/// If the array is empty, returns empty string
```
Keats: might be worth changing all of those to return nil - but not in this PR

Keats:
The v1 branch has one failing test, don't worry about it. If you have any experience with impl trait or similar maybe you can help there too: https://github.com/Keats/tera/pull/355#issuecomment-439999866

Keats:
I think you will need to rebase for the v1 branch, we don't have error-chain anymore for example

RealOrangeOne:
ok, i did terrible things to the git history, and completely messed the history. Gunna re-do my changes on a fresh clone of a fresh fork

PR #362, opened 2018-11-26
fix test

andy128k:
Borrowing doesn't work because of `Arc::new` inside of `register_function`. It requires ownership.
This commit fixes a test and shows how to have a function with shared state.


Keats:
I see, thanks a lot!

PR #363, opened 2018-11-26
Add `nth` filter

### LINKED ISSUE #358: Add nth filter

RealOrangeOne:
Fixes #358 

Re-implements https://github.com/Keats/tera/pull/361 because some days i'm bad at git.


Keats:
Thanks!

PR #378, opened 2019-01-13
WIP. Lifetime

andy128k:
Related to https://github.com/Keats/tera/issues/299

### DIFF
--- a/examples/actix/src/main.rs
+++ b/examples/actix/src/main.rs
@@ -8,7 +8,7 @@ use actix_web::{
 
 
 struct AppState {
-    template: tera::Tera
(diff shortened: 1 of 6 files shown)


Keats:
Thanks, I'll try it!
It is a bit worse in terms of ergonomics but if it works, it is a huuuge perf win compared to the current state.

Keats:
I had a quick look in a flight yesterday but I don't think this will work. Rather than a reference, the functions need to have a lifetime bound other than `'static'. Afaik, the test passes because they share the scope

Keats:
In the end, `Arc<RwLock<_>>` does the job, no need for lifetimes

PR #398, opened 2019-04-19
API enhancement proposals

jimmiebfulton:
Hi,

I'm using Tera to create a tool similar to Maven's Archetypes.  Like cargo generate, but much more flexible.  Thanks.

I'm essentially using a template directory, whereby the directories and file names themselves can contain variables, and the files contained within are contents for which variable replacement should occur.

I'm finding it very appealing to have the ability to set up a Tera instance that is essentially using each file in the template directory as a bunch of one-off templates.  With the current API, I'm restricted to either using a one_off Tera instance, or adding each file as a template itself, even though it will only be rendered once.

I have changed the render method to accept anything that can be turned into a Value.  This may cut down on having additional 'overload' methods for various types.  I didn't spend the time to see how perhaps if there is a way to `impl From<Serializable> for Value`, which would allow the removal of the render_value method.  I don't know if you had already considered this approach and decided it was too opaque for end users.  I also added render_template and render_string methods, and delegate to those from render itself.  I haven't spent any time testing any of this out, other than running existing unit tests.

This is more of a proposal than a pull request.  If you are interested in exploring these changes, I'd be happy to spend some more time on tests and docs, unless you wanted to take this and do whatever you'd like, if anything.

Cheers,

Jimmie

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -132,6 +132,16 @@ impl ValueTruthy for Value {
     }
 }
 
+impl From<Context> for Value {
+    fn from(ctx: Context) -> Self {
+        let mut m = Map::new();
+        for (key, value) in ctx.data {
+            m.insert(key, value);
+        }
+        Value::Object(m)
+    }
+}
+
 /// Converts a dotted path to a json pointer one
 #[inline]
 pub fn get_json_pointer(key: &str) -> String {
--- a/src/context.rs
+++ b/src/context.rs
@@ -48,11 +48,7 @@ impl Context {
 
     /// Converts the context to a `serde_json::Value` consuming the context
     pub fn into_json(self) -> Value {
-        let mut m = Map::new();
-        for (key, value) in self.data {
-            m.insert(key, value);
-        }
-        Value::Object(m)
+        self.into()
     }
 }
 
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -17,6 +17,7 @@ use crate::errors::{Error, Result};
 use crate::renderer::Renderer;
 use crate::template::Template;
 use crate::utils::escape_html;
+use serde_json::Value;
 
 /// The escape function type definition
 pub type EscapeFn = fn(&str) -> String;
@@ -301,10 +302,9 @@ impl Tera {
     /// // Rendering a template with an empty context
     /// tera.render("hello.html", Context::new());
     /// ```
-    pub fn render(&self, template_name: &str, context: Context) -> Result<String> {
+    pub fn render<C: Into<Value>>(&self, template_name: &str, context: C) -> Result<String> {
         let template = self.get_template(template_name)?;
-        let renderer = Renderer::new(template, self, context.into_json());
-        renderer.render()
+        self.render_template(&template, context)
     }
 
     /// Renders a Tera template given an object that implements `Serialize`.
@@ -328,6 +328,37 @@ impl Tera {
         let template = self.get_template(template_name)?;
         let renderer = Renderer::new(template, self, value);
         renderer.render()
+    }
+
+    /// Renders a one off template (for example a template coming from a user input) given a `Context`
+    ///
+    /// ```rust,ignore
+    /// // Rendering a template with a normal context
+    /// let mut context = Context::new();
+    /// context.insert("subject", "world");
+    ///  let template = Template::new("one_off", None, contents).unwrap();
+    /// tera.render_template(&template, context);
+    /// ```
+    pub fn render_template<C: Into<Value>>(&self, template: &Template, context: C) -> Result<String> {
+        let render = Renderer::new(&template, self, context.into());
+        render.render()
+    }
+
+    /// Renders a one off string (for example a string template coming from a user input) given a `Context`
+    ///
+    /// Any errors will mention the `one_off` template: this is the name given to the template by
+    /// Tera
+    ///
+    /// ```rust,ignore
+    /// // Rendering a template with a normal context
+    /// let mut context = Context::new();
+    /// context.insert("subject", "world");
+    /// tera.render_one_off("Hello, {{ subject }}", context);
+    /// ```
+    pub fn render_string<C: Into<Value>>(&self, template_contents: &str, context: C) -> Result<String> {
+        let template = Template::new("one_off", None, template_contents)
+            .map_err(|e| Error::chain(format!("Failed to parse template string: '{}'", template_contents), e))?;
+        self.render_template(&template, context)
     }
 
     /// Renders a one off template (for example a template coming from a user input) given a `Context`
@@ -749,7 +780,7 @@ mod tests {
             ("c", "{% extends \"d\" %}"),
             ("d", ""),
         ])
-        .unwrap();
+            .unwrap();
 
         assert_eq!(
             tera.get_template("a").unwrap().parents,
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -8,6 +8,7 @@ use std::sync::Arc;
 use globwalk::glob;
 use serde::Serialize;
 use serde_json::value::to_value;
+use serde_json::Value;
 
 use crate::builtins::filters::{array, common, number, object, string, Filter};
 use crate::builtins::functions::{self, Function};
@@ -17,7 +18,6 @@ use crate::errors::{Error, Result};
 use crate::renderer::Renderer;
 use crate::template::Template;
 use crate::utils::escape_html;
-use serde_json::Value;
 
 /// The escape function type definition
 pub type EscapeFn = fn(&str) -> String;
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -780,7 +780,7 @@ mod tests {
             ("c", "{% extends \"d\" %}"),
             ("d", ""),
         ])
-            .unwrap();
+        .unwrap();
 
         assert_eq!(
             tera.get_template("a").unwrap().parents,


jimmiebfulton:
Interesting.  I see you've already built (kickstart) exactly what I was building.  Support for variables in directories and files.  Even the same [[variables]] section with identical names for the fields.  Exactly what I wanted.  Thanks!

Keats:
>  With the current API, I'm restricted to either using a one_off Tera instance, or adding each file as a template itself, even though it will only be rendered once.

What's the issue with using https://docs.rs/tera/1.0.0-beta.4/tera/struct.Tera.html#method.one_off_value ?

I like the `From` idea though.

jimmiebfulton:
The problem with using one_off or one_off_value is exactly what the documentation states regarding these two, in that you don't get a built-up Tera instance with all of the functions and Templates you'd like to have available.  I guess my ideal would be able to have my cake and eat it too :-), where by a Tera instance can have all of the functions and templates registered, and any "one off" templates can simply make use of this existing instance.  I don't know anything about Jinga, and what I've learned so far is purely through Tera.  It would would seem nice that a "one off" template could still be able to make use of inheritance, without itself being registered as a template that could serve as a parent.  Essentially, a middle ground.  If you register a Template, it can be included, or used as a parent, but if you just do a one-off, it can't itself be included or serve as a parent, but could make use of other templates and use them for include and as parents, and also make use of all the functions made available on the proper Tera instance.  I hope this makes sense.

The use case I was going for is exactly what is going on in kickstart.  I haven't looked at kickstart code, but my approach prior to my discovery of kickstart would be that I have two directories.  One directory contains the templated project files, and are each essentially one-offs.  Another directory is templates, containing templates that can either be included, or serve as parent templates for the projects files.  I'd register a number of filters, including 'snake_case', 'train_case', 'pascal_case', 'camel_case', all of which can be used to take a single variable, and have it represented in different contexts.  Example: you are prompted a 'Service Name', which may be typed in as 'Order Service'.  In Java package names, this might show up as `com.{{ company }}.{{ name | package_case }};`  resulting in com.company.order.service.  The class name might be `public class {{ name | pascal_case }}Impl`.  As a field name: `private {{ name | pascal_case }}Client {{name | camel_case }}client;`. The templates could contain something like a BaseJava.java file containing common header (licensing, etc) information, and any file in the main project files serving as one-offs could make use of these filters and the BaseJava.java file, perhaps eliminating a bunch of duplicated information in templates.

I've typically done this with Maven Archetypes, which uses Velocity templates, and have been missing this more broadly in other language, not being aware of cookiecutter, and now kickstart.  Maven Archetypes are interesting in that they are published as JAR files in Artifactory or Nexus, etc.  But otherwise, building them is a real pain.  The cookiecutter/kickstart approach is a lot simpler in maintaining, and frankly, more powerful, even if you give up a few Java-specific features.

I hope all this makes sense, and sorry for the long message on a PR.  Just giving some context.  Thanks for taking a peek at it.

Keats:
I see what you mean. Ideally you would instantiate one instance of Tera, add whatever filters/functions you need like the casing ones. I *think* you can get away with not having inheritance but I'm not familiar at all with Maven/Java.
It would make sense to have a `render_string` method on the Tera instance though to be able to reuse a custom Tera instance easily though. You can currently do that by having a small function creating a default Tera instance with your custom filters but it's a bit cumbersome and would be easier if it is built-in.

I'd be happy to add the cases filters to `kickstart` for example if you think it can be used!

Keats:
Sadly it looks like the `Into` approach won't work since what we actually use is `Serialize`, `to_value` can fail and `TryInto` was just stabilized recently. I could be wrong though

What do you think of https://github.com/Keats/tera/commit/cf236ca655a83c8ede61fb5ee16a00bb616b3e81 for the render string method?

jimmiebfulton:
Hi,

I thought I’d address this offline.  I think that cf236ca would work.

I keep coming across your name in Rust world, and I like the work you’ve been doing.  As I indicated before, I wanted something just like kickstart and began one of my own, which is how I came across Tera.  Then, I discovered Kickstart, too.  So now I’m a fan of two of your projects.

I’ve started building my own template engine as a way to learn Pest, and perhaps Nom, for parsing, and I’ve been going through the Tera code.  I’d like to experiment with some potential API enhancements.  Perhaps I could share my progress with you, and see if there are any ideas you might want to incorporated into Tera.  I was considering applying a more Rust-like syntax instead of a Python-like syntax.

{%
for i in (0..10) {
%} Iteration: {{ i }} {%
}
%}

Or

{% for (i, person) in people.enumerate() { %}
<tr><td>{{ i }}></td>{{ person.lname }}, {{ person.fname }}<td></td></tr>
{% } %}

I dunno.  Could end up being ugly, but I’d like to see how it goes.  At the least, it’s a learning experience for me.  It would seem I’d need to tap into Syn for Rust AST.

I’m also interested in experimenting with the ability to have templates that load from the file system on every render, allowing for developers to tweak templates without restarting (maybe I’m just too used to slow Java start times), but also allow them to potentially compile the AST into the Rust binary for releases.  Kinda like the rust-embed library: https://github.com/pyros2097/rust-embed.

One approach would be too create different implementations of a TemplateRepository trait.  DiskTemplateRepository, PostgresTemplateRepository, WebDavTemplateRepository, which support the ability to be wrapped by a CompilingTemplateRepository which eagerly loads all the templates and writes their resulting AST into the binary.  Just a rough sketch.  I like Tera for allowing me to dynamically load templates, and I like the idea of Askama for high-performance compiled templates.  I think its possible to get the best of both, and would make for a killer template engine.


As far as Kickstart goes, I think it’s almost exactly what I want, but I’d also like to experiment with some variations on the concept as well.

For instance.  I’m adding the ability to incorporate “answer files”.  You could have an answer file in your home directory with answers to variables that frequently occur in project templates.  For instance, I could set author to ‘Bob <bob@alice.coml>’, so that any templates that use the author convention would simply respect my answer.  However, I’d also like to allow template authors to include answer files in their templates, with the intention of supporting ‘companion’ templates.

Let’s say we have a template for creating a project.  It’s a multi-module project containing API, Client, Impl, and Persistence modules.  There is also a template that allows for the ability to create new entities.  If you were in the root of your generated project, and ran kickstart in that directory using the “add new entity” template, it could create a new Customer in API, a CustomerEntity and CustomerRepository in persistence, etc.  However, you wouldn’t want to have the user be required to answer some of the same questions again.  Therefore, the template author can also include an anwser file template, which simply fills in the variables for the answers that should be saved to the project.  kickstart would look for either a kickstart.toml or .kickstart.toml, perhaps in that order, and use those answers to apply new objects into a project.

So ~/.kickstart/answers.toml:
[[answers]]
name=“author”
answer=“Bob <bob@alice.com>”

In ~/.kickstart/templates/.answers.toml:
[[answers]]
name=“package”
answer=“{{ package }}”

[[answers]]
name=“name”
answer=“{{ name }}”


Which would generate in ~/projects/customer-service/.answers.toml:
[[answers]]
name=“package”
answer=“com.foo.customer.service”

[[answers]]
name=“name”
answer=“Customer Service”

Now, the uses can apply new “sub templates” designed to work with the same structure as the template that created the project in the first place, and the user doesn’t need to supply their author for new projects they create, or re-type “com.foo.customer.service”, etc.  The template author gets to decide which answers should be persisted and re-used on subsequent runs.

In addition, I’ve got the idea of catalog.toml, whereby a Catalog has a list of template paths and other metadata.  It could point to a directory, zip or tarball, or a Git repo.  Therefore, a user could do something like:

kickstart add —catalog java “http://github.com….”
kickstart add —catalog python ~/.kickstart/templates/flask

This gives user the ability to create their own menus of templates as they see fit.  They can refer to Github projects, if they just want to follow master, or they can check it out and fork changes locally.

kickstart create —catalog java
1) Spring Boot, Multi-module Project
2) Spring Boot, Multi-module Add Entity
3) Guice, Library

Lots of ideas I’d like to explore, as I use “Archetypes” all the time in professional work, and really want to shave down the time for bootstrapping and working in projects.

Like I said, I’d be happy to share my progress, and you’re welcome to add these concepts to your own projects.  I’m learning a lot from yours.

I can see some of the proposals for templates for cargo new, but what I’ve seen so far are kinda convoluted.  I like kickstart the best, but it may be too sophisticated, and general-purpose, for cargo new.  However, I think it’s what many might want; I certainly do.  It’s completely possible to develop kickstart as a more sophisticated replacement for cargo new for those that want more power.  Unfortunately, I don’t know how many people have looked at kickstart and appreciate its flexibility.

Feel free to keep in touch.  I don’t have experience contributing to open source, but I’m inclined to start contributing to Rust projects, and your projects are what’s current drawn my interest. Happy to collaborate.

Keats:
> I’m also interested in experimenting with the ability to have templates that load from the file system on every render, allowing for developers to tweak templates without restarting (maybe I’m just too used to slow Java start times), but also allow them to potentially compile the AST into the Rust binary for releases. 

If you want to use Rust logic in the template that's going to be tricky. You will need to re-implement a good chunk of Rust grammar/logic (hell maybe even the borrow checker) to have it working as plain files. If it is just mimicking the Rust logic then it is different but you might run into issues when compiled to Rust code with borrowing.
That was an initial goal for Tera but it is way too much work imo since you're not going to end up with something as nice as a interpreted template engine or a compiled template engine.

As for the kickstart comments, those are interesting! I didn't think of sub templates since tools I use usually have their own cli that thandle them, eg `manage.py` in django to create a new app and at least in my part of the programming world, no one would use a tool like kickstart for that. Looking at https://github.com/audreyr/cookiecutter#a-pantry-full-of-cookiecutters it's all about setting the initial state of the project. I like the idea of catalogs, I probably don't use templates enough to be worth it but it seems like it would be a nice addition.

> I can see some of the proposals for templates for cargo new, but what I’ve seen so far are kinda convoluted.

Yes they are way too basic for actual use, I don't really see when I would use cargo new personally.

> Feel free to keep in touch.  I don’t have experience contributing to open source, but I’m inclined to start contributing to Rust projects, and your projects are what’s current drawn my interest. Happy to collaborate.

Happy to do so and don't hesitate if you have questions.

PR #399, opened 2019-04-28
[WIP] fuzzing fixes

stusmall:


sntdevco:
cc @Keats

Keats:
I've seen it, I need to see if we can prevent that state in the grammar itself. I don't know when I'll have time for it though

Keats:
Done in the v1 branch

PR #400, opened 2019-05-01
Add object builtin tester.

sennav:
Hello I missed a builtin tester to check if a variable in context can be iterated over key value (if it is a hash). I made a local implementation of it in my project but this seems like something more people can use :)

I'm a rust beginner so let me know if there's anything I can improve to get this merged.


Keats:
Hi can you re-do this PR on the v1 branch instead? The code itself looks fine but will need to change with the new api in v1

sennav:
Sure, I'll close this one then and come back with the correct one.

PR #401, opened 2019-05-02
Add object builtin tester.

sennav:
Much better with the tests passing now :)

### DIFF
--- a/docs/content/docs/templates.md
+++ b/docs/content/docs/templates.md
@@ -897,6 +897,9 @@ Example:
 #### iterable
 Returns true if the given variable can be iterated over in Tera (ie is an array/tuple).
 
+#### object
+Returns true if the given variable is an object (ie can be iterated over key, value).
+
 #### starting\_with
 Returns true if the given variable is a string starts with the arg given.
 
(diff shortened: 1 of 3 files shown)


docs/content/docs/templates.md
```diff
@@ -897,6 +897,9 @@ Example:
 #### iterable
 Returns true if the given variable can be iterated over in Tera (ie is an array/tuple).
 
+#### object
+Returns true if the given variable is a hash (ie can be iterated over key, value).
```
Keats: is a hash -> is an object
A hash would be a string :)

sennav: Done

Keats:
Thanks!

sennav:
Thanks, since I got in touch here do you mind to answer a question really quick? I'm trying to use tera for customising output from a CLI tool (maybe this is not the main goal of tera, but anyway) and when I try to print the escape character Tera just escapes the initial '\' into '\\'. I've tried to replace the escape function but that seems to be used only for variables (I'm trying to do a macro that can make text bold for instance). Can I make a PR to allow that? Would you point me in the direction of where it is being escaped (I've looked through the code but I didn't figured it out yet), thanks.

Keats:
The initial what?

sennav:
Oh sorry I didn't checked the preview before sending. I'm trying to create this macro:
```
{%- macro bold(t) -%}
{%- set bold = '\x1b[7m' -%}
{%- set end = '\x1b[0m' -%}
{{ bold }}{{ t }}{{ end }}
{%- endmacro bold -%}
```
and `\x1b[7m` gets transformed into `\\x1b[7m`.

PR #403, opened 2019-05-13
Add custom singular suffix for pluralize filter

### LINKED ISSUE #402: Pluralize should take an optional prefix
Some more complex words like 'category' can't be pluralized using the current `pluralize` filter. It should do what django does and accept an optional prefix so that you can do something like:

	categor{{ 1 | pluralize(prefix="y", suffix="ies") }}

which results in:

	category

64:
Fixes #402.


Keats:
Thanks!

PR #426, opened 2019-07-05
Implemented feature from issue #395 ("else" case in for loops)

### LINKED ISSUE #395: Add support for Django's {% empty %} in for-loops
I am missing support for an `{% empty %}` statement in for-loops. I know the [`{% empty %}` statement](https://docs.djangoproject.com/en/2.1/ref/templates/builtins/#for-empty) from Django's templating language:

> The for tag can take an optional {% empty %} clause whose text is displayed if the given array is empty or could not be found:
> 
> ```
> <ul>
> {% for athlete in athlete_list %}
>     <li>{{ athlete.name }}</li>
> {% empty %}
>     <li>Sorry, no athletes in this list.</li>
> {% endfor %}
> </ul>
> ```

MathiasPius:
Fixes #395 

I've implemented the functionality requested in issue #395 along with some test cases that cover the code.

I've made the following decisions that I'd like some feedback on:

### Renamed {% empty %} -> {% else %}
I've taken the liberty of naming the fail-case "else" instead of "empty" since I feel like it better conveys the control flow, "empty" feels a little ambiguous, at least as someone not familiar with Django. 

### Silently render _else_ case if iterating over undefined variable
As requested, this functionality also covers the case where you attempt to iterate over an undefined variable, defaulting to the _else_ block, and throwing an error if no _else_ block is defined.

This will fail to compile:
```
{% for a in undefined_variable %}
    {{a}}
{% endfor %}
```
and yield the error:  `Tried to iterate undefined container 'undefined_variable'. To treat undefined containers the same as empty ones, explicitly define an {% else %} block within the for loop`

This will render "hello world"
```
{% for a in undefined_variable %}
    {{a}}
{% else %}
    hello world
{% endfor %}
```

I'm unsure of whether this is a good compromise, and if the "undefined" case should even be covered by the _else_ case, or if additional checks by the user should be required in that case.

### DIFF
--- a/src/renderer/processor.rs
+++ b/src/renderer/processor.rs
@@ -172,17 +172,13 @@ impl<'a> Processor<'a> {
         let for_loop_body = &for_loop.body;
         let for_loop_empty_body = &for_loop.empty_body;
 
-        let container_val = if let Ok(val) = self.safe_eval_expression(&for_loop.container) {
-            val
-        } else {
-            if let Some(empty_body) = for_loop_empty_body {
-                return Ok(self.render_body(&empty_body)?);
-            } else {
-                return Err(Error::msg(format!(
-                    "Tried to iterate undefined container `{}`. To treat undefined containers the same as empty ones, explicitly define an {{% else %}} block within the for loop",
-                    container_name,
-                )));
-            }
+        let container_val = match (self.safe_eval_expression(&for_loop.container), for_loop_empty_body) {
+            (Ok(container_val), _) => container_val,
+            (_, Some(empty_body)) => return Ok(self.render_body(&empty_body)?),
+            (_, _) => return Err(Error::msg(format!(
+                "Tried to iterate undefined container `{}`. To treat undefined containers the same as empty ones, explicitly define an {{% else %}} block within the for loop",
+                container_name,
+            )))
         };
 
         let for_loop = match *container_val {
--- a/src/renderer/processor.rs
+++ b/src/renderer/processor.rs
@@ -172,6 +172,7 @@ impl<'a> Processor<'a> {
 
         let for_loop_name = &for_loop.value;
         let for_loop_body = &for_loop.body;
+        let for_loop_empty_body = &for_loop.empty_body;
 
         let for_loop = match *container_val {
             Value::Array(_) => {
@@ -210,6 +211,14 @@ impl<'a> Processor<'a> {
         };
 
         let len = for_loop.len();
+        if len == 0 {
+            if let Some(empty_body) = for_loop_empty_body {
+                return Ok(self.render_body(&empty_body)?);
+            } else {
+                return Ok("".to_string());
+            }
+        }
+
         self.call_stack.push_for_loop_frame(for_loop_name, for_loop);
 
         let mut output = String::with_capacity(len * 20);
--- a/src/renderer/processor.rs
+++ b/src/renderer/processor.rs
@@ -172,14 +172,7 @@ impl<'a> Processor<'a> {
         let for_loop_body = &for_loop.body;
         let for_loop_empty_body = &for_loop.empty_body;
 
-        let container_val = match (self.safe_eval_expression(&for_loop.container), for_loop_empty_body) {
-            (Ok(container_val), _) => container_val,
-            (_, Some(empty_body)) => return Ok(self.render_body(&empty_body)?),
-            (_, _) => return Err(Error::msg(format!(
-                "Tried to iterate undefined container `{}`. To treat undefined containers the same as empty ones, explicitly define an {{% else %}} block within the for loop",
-                container_name,
-            )))
-        };
+        let container_val = self.safe_eval_expression(&for_loop.container)?;
 
         let for_loop = match *container_val {
             Value::Array(_) => {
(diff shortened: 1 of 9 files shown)


Keats:
> I've taken the liberty of naming the fail-case "else" instead of "empty"

Good for me.

> As requested, this functionality also covers the case where you attempt to iterate over an undefined variable, defaulting to the else block, and throwing an error if no else block is defined.

I think this is ok, I just need to think about it for a bit. I'm like 60% ok with that but 40% of me still think it should error even with an else block.

Don't worry about the CI failure, looks like a sub-subdependency updated their minimum version of Rust...

src/renderer/processor.rs
```diff
@@ -210,6 +222,14 @@ impl<'a> Processor<'a> {
         };
 
         let len = for_loop.len();
+        if len == 0 {
+            if let Some(empty_body) = for_loop_empty_body {
+                return Ok(self.render_body(&empty_body)?);
+            } else {
+                return Ok("".to_string());
+            }
```
Keats: I think a `match` here and on the similar if/else block above would be neater but very minor

MathiasPius: Good point, rewrote it using compound match statements, not sure if it's made it clearer what's going on so feel free to rework it :)

I agree with your reservations regarding the "undefined" case. Actually, couldn't you remove it and simply cover that special case with something like this:
```
{% for a in undefined_variable | default(value=[]) %}
    {{ a }}
{% else %}
    hello world
{% endfor %}
```

Keats: It does work, the only issue is more about the principle of least surprise: an undefined condition evaluates is falsy so someone could expect an undefined contained to be falsy/empty and go into the `else` like jinja2 does.

MathiasPius: I can definitely see that. I guess the question boils down to how closely tera is meant to emulate Jinja2. Since it is a Rust project, I suppose could be argued that somewhat explicit typing and error-handling is preferred. Of course the end-user will likely neither know nor care about that fact.

It's your call to make, but if you land on emulating Jinja2 in this case, I think we should stick with the original naming of the empty-case block and use {% empty %}, just so there are no surprises.

awulkan:
> I think this is ok, I just need to think about it for a bit. I'm like 60% ok with that but 40% of me still think it should error even with an else block.

I'm also on the fence about this. It could introduce unexpected behavior without you knowing about it. On the other hand, it appears that Django does just that, and if it works for them it should work for us, right?

>The for tag can take an optional {% empty %} clause whose text is displayed if the given array is empty **or could not be found**
https://docs.djangoproject.com/en/2.1/ref/templates/builtins/#for-empty

Keats:
The main difference between Tera and Jinja2/Django is that it does error when using undefined variables outside of conditions. `{{ undefined_var }}` will be a runtime error in Tera but works just fine in Jinja2.

AFAIK the only exception to errors in Tera for undefined is `{% if undefined_var %}` which evals to `false` since it can't get painful in some templates otherwise

MathiasPius:
In that case, I think going with the PR in its current state, where undefined containers trigger hard errors regardless of the presence of {% else %} blocks is the way to go.

Keats:
Let's go with that PR and we can change it later if needed. Thanks!

jplatte:
Not sure if this can still be changed back to `empty`, but I think it would definitely be worth it because this does the exact opposite of Python's `for ... else` using the exact same keyword, and I don't think there's precedent anywhere else for doing it like this. On top of being confusing syntax in the first place, as is the general consensus in the Python community AFAIK, this just makes it more confusing for tera users.

([Python's else branch executes if the loop completed normally](https://book.pythontips.com/en/latest/for_-_else.html#else-clause))

Keats:
This is following the Jinja2 syntax: https://jinja.palletsprojects.com/en/2.10.x/templates/#for but it might just be legacy https://github.com/pallets/jinja/issues/77

I personally prefer `empty` since the else in Python is super confusing 
cc @MathiasPius

jplatte:
Oh wow, a python templating library uses the same keyword as python to basically mean the opposite thing :sweat_smile: 

> Note that, in Python, *else* blocks are executed whenever the corresponding loop **did not** *break*. Since Jinja loops cannot *break* anyway, a slightly different behavior of the *else* keyword was chosen.

"slightly different"...

@Keats I guess since this PR has been merged a few months ago, renaming this construct in tera would be a breaking change, correct?

Keats:
>  guess since this PR has been merged a few months ago, renaming this construct in tera would be a breaking change, correct?

No, it's in the v1 beta so it can have breaking changes without issues still.

MathiasPius:
I was not aware of the use-case in Python prior to designing this, but I would suggest that using an adjective as a control structure element makes very little sense, since it is entirely unclear what it applies to. I also think Python's use of the word `else` in this case is very misleading, and should instead have been `then` or similar.

The choice to use `else` was, as @Keats mentioned, based on what Jinja2 does, and I still personally prefer it, but have no issue with changing it if using `empty` is the more common term for this structure. I also wouldn't mind having some more examples of other template engines using `empty` however. A cursory search of other engines I could remember hearing about like handlebars or twig, showed that handlebars didn't seem to have the structure at all, and Twig used `else`.

jplatte:
Seems like `for` ... `else` with the meaning this PR gives it actually has a lot of precedent in templating languages. I was only aware of Python doing the opposite, but maybe it's fine to keep this afterall. I've found:

* [Liquid has for ... else](https://shopify.github.io/liquid/tags/iteration/) (this is the templating engine Jekyll uses)
* [mustache](https://mustache.github.io/mustache.5.html) doesn't have control flow, so no for ... else or equivalent
* [Nunjucks has for ... else](https://mozilla.github.io/nunjucks/templating.html#for)
* [Pug has for ... else](https://pugjs.org/language/iteration.html)

PR #427, opened 2019-07-05
Fix documentation for escape

makuk66:


Keats:
Thanks!
I've pushed the change on the v1 branch as the docs had a small update and fixing conflicts would have harder than changing the character.

PR #430, opened 2019-07-09
Add whitespace control to expressions

### LINKED ISSUE #429: Support removing whitespace around expressions
I had some html like this:

```html
<li><a class="p-category"
  href="{{ tag_url }}">
  {{ tag }}
</a></li>
```
This produces a space that is underlined because the whitespace is included.

Screen Shot of link with trailing whitespace underlined

Removing the spaces in the template removes the trailing whitespace.

```html
<li><a class="p-category"
  href="{{ tag_url }}">{{ tag }}</a></li>
```

Screen Shot of link with trailing whitespace removed

Since whitespace removal for statements is supported using `{%-` and `-%}`, could a similar operator for expressions be added? In my case the whitespace wasn't hard to remove manually, but it seemed like a missing feature.

Geemili:
Adds alternate start and end tokens for expressions (`{{-` and `-}}`) that allow the user to remove whitespace around expressions.

Closes #429

### DIFF
--- a/src/parser/ast.rs
+++ b/src/parser/ast.rs
@@ -317,7 +317,7 @@ pub enum Node {
     /// Some actual text
     Text(String),
     /// A `{{ }}` block
-    VariableBlock(Expr),
+    VariableBlock(WS, Expr),
     /// A `{% macro hello() %}...{% endmacro %}`
     MacroDefinition(WS, MacroDefinition, WS),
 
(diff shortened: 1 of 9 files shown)


Keats:
Thanks for that! It looks like the error messages tests need to be updated strangely

Geemili:
It probably is because I had to make the variable start and end tokens visible. Should I just go ahead and the tests?

Keats:
Please do!

Geemili:
Okay, the error messages are fixed

Keats:
Thanks!

PR #441, opened 2019-08-24
array map filter implementation

tones111:


Keats:
That looks neat, thanks. I'm a bit worried about using the same name as in Jinja2 but without the fact that we can call arbitrary functions like lower, upper though.

How about adding a function argument with a few built-ins like `lower` & `upper`?

tones111:
I think implementing that functionality for only a couple of filters would cause confusion when users try to figure out why some work and some don't.  If this were to be implemented I think the filter API would need a way to access the Tera instance so it could call get_filter.  I couldn't see a way to do that currently, but am very new to this library and might be overlooking something.

tones111:
Implementing the filter functionality within map significantly complicates the implementation.  IMHO I believe a better approach would be to keep the behaviors in separate filters (would still need to provide the Tera object to the filter).  Since the filters don't receive non-keyword args we can't match the jinja2 specification anyway.

Please let me know which direction you prefer.

Keats:
I'm aware we cannot match the Jinja2 syntax but to me map is more about applying functions than plucking members

> IMHO I believe a better approach would be to keep the behaviors in separate filters (would still need to provide the Tera object to the filter).

What do you mean?

tones111:
The behavior of map(attribute="abc") doesn't share much (anything?) with map(filter="upper") so trying to implement them within the same filter adds unnecessary complexity.  The primary motivation for it is consistency with jinja2's implementation, and perhaps that's a compelling enough reason.

It would make more sense to me to implement each behavior in its own filter.  So there would be something like...
map(attribute="abc") which would extract the inner field (similar to rust's iterator map)
filter(func="upper") which allows non-array filters to be used with an array input.

There's also the question of how to handle errors that arise when applying the filter against the array.  Should processing stop and return the error, or should it take a best-effort approach and process what it can.  I implemented the second approach (with test) but it wouldn't be hard to flow the error up instead.

Keats:
I think the last commit won't work, filters can't take a Tera instance as parameter. Let's revert the last commit and I'll play with it a bit with functions when I have time.

> The behavior of map(attribute="abc") doesn't share much (anything?) with map(filter="upper") so trying to implement them within the same filter adds unnecessary complexity

I agree. It's just that in programming `map` is used to implement both so it's pretty compelling. In practice for a template engine though, calling a function on all members of an array is not *super* useful so we can probably skip it, at least initially.

Keats:
Sorry I forgot to mention: any chance you can update the docs as well? In `docs/contents/docs/_index.md`, adding it to the list of builtin filters

Keats:
Thanks!

PR #443, opened 2019-08-24
Error can be passed across threads (fixes #442)

### LINKED ISSUE #442: tera::Error cannot be sent between threads safely
Hi, i'm using error_chain in a program that does tera templating in a rayon loop. Updating from v0.11 to v1.0.0-beta14 produced the following error:

```
error[E0277]: `(dyn std::error::Error + 'static)` cannot be sent between threads safely
  --> src/main.rs:21:1
   |
21 | / error_chain!{
22 | |     foreign_links {
23 | |         Http(reqwest::Error);
24 | |         Json(serde_json::Error);
...  |
27 | |     }
28 | | }
   | |_^ `(dyn std::error::Error + 'static)` cannot be sent between threads safely
   |
   = help: the trait `std::marker::Send` is not implemented for `(dyn std::error::Error + 'static)`
   = note: required because of the requirements on the impl of `std::marker::Send` for `std::ptr::Unique<(dyn std::error::Error + 'static)>`
   = note: required because it appears within the type `std::boxed::Box<(dyn std::error::Error + 'static)>`
   = note: required because it appears within the type `std::option::Option<std::boxed::Box<(dyn std::error::Error + 'static)>>`
   = note: required because it appears within the type `tera::errors::Error`
   = note: required because it appears within the type `ErrorKind`
   = note: required because it appears within the type `Error`
   = note: this error originates in a macro outside of the current crate (in Nightly builds, run with -Z external-macro-backtrace for more info)

error: aborting due to previous error

For more information about this error, try `rustc --explain E0277`.
```

Full extract from my source file:
```
error_chain!{
    foreign_links {
        Http(reqwest::Error);
        Json(serde_json::Error);
        Io(std::io::Error);
        Templating(tera::Error);
    }
}
```

southerntofu:
Hi, this solves my problem in #442.

I'm not aware if removing this line breaks things for other people, but it does not affect any test.

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -1,8 +1,6 @@
 use serde_json;
 
 error_chain! {
-    errors {}
-
     foreign_links {
         Json(serde_json::Error) #[doc = "An error happened while serializing JSON"];
     }


Keats:
Any PR should be done on the v1 branch rather than master which doesn't even build anymore

PR #445, opened 2019-08-29
array unique filter implementation

tones111:
This change implements the ability to filter out duplicates from arrays for some types. The filter supports a case insensitive comparison that is implemented via the simplistic to_lower String method. Fancier behavior would require external dependencies.


src/builtins/filters/array.rs
```diff
@@ -91,6 +91,61 @@ pub fn sort(value: &Value, args: &HashMap<String, Value>) -> Result<Value> {
     Ok(sorted.into())
 }
 
+/// Remove duplicates from an array.
+/// Use the 'attribute' argument to define a field to filter on.
+/// For strings, use the 'case_sensitive' argument (defaults to false) to control the comparison.
+pub fn unique(value: &Value, args: &HashMap<String, Value>) -> Result<Value> {
+    use std::mem::discriminant;
```
Keats: :o I didn't know that. Since it's only used twice can you prefix instead so it's easy to see where it's coming from?

Keats:
Thanks!

PR #447, opened 2019-09-02
indent filter implementation

tones111:
This change implements the [indent](https://jinja.palletsprojects.com/en/2.10.x/templates/#indent) filter.  Unfortunately I couldn't find a good way to preserve the input line ending, but maybe that's less important these days.

### DIFF
--- a/docs/content/docs/_index.md
+++ b/docs/content/docs/_index.md
@@ -778,6 +778,9 @@ By default, the filter will add an ellipsis at the end if the text was truncated
 change the string appended by setting the `end` argument.
 For example, `{{ value | truncate(length=10, end="") }}` will not append anything.
 
+#### indent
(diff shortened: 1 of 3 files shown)


Keats:
What's the usecase for that filter? I can only think of YAML templating and that is niche enough that endusers can implement that themselves if needed. Is there another usecase?

tones111:
The main use case I was thinking of is to assist when generating source code.  Often times you'll have a reusable block of text (from a function or macro) that needs to be indented by varying amounts depending on the surrounding context.

The other use case is trying to achieve feature parity with other jinja2 implementations.  Adoption is likely to be more challenging if there aren't functional equivalents to accomplish what's available elsewhere.  If the concern is about code bloat I suppose these additions could be hidden behind a feature flag.  A stronger argument could be made if the functionality in question pulled in large dependencies.

Keats:
> The other use case is trying to achieve feature parity with other jinja2 implementations. 

This is not a Jinja2 implementation though, it is only inspired by it. In the case of the indent filter, I would wait to see if people actually need it before implementing it rather than copy all of the Jinja2/Django filters.

Keats:
This filter is not going to be added for now.

untereiner:
Hi,

Could this PR related to https://github.com/Keats/tera/issues/670 ?
I would also need to indent generated text.

Issue #434, opened 2019-07-11
Add String to Number Filter


naftulikay:
In my [Tera rendering CLI](https://github.com/naftulikay/jinjer), I expose environment variables to the Tera context. Environment variables, of course, are inherently strings, and this makes trying to pass integers/floats not possible:

```jinja
{{ env.ATTEMPT_COUNT + 1 }}
```

`ATTEMPT_COUNT` is set to `"3"` as a string. This template fails to render because adding one to a string doesn't make any sense. I don't see any filters presently for type conversions from strings to number types.

I'd ultimately like to do something like this:

```jinja
{{ (env.ATTEMPT_COUNT | to_u32) + 1 }}
```

I would like to contribute to this feature but my availability is low right now.

Keats:
Yeah there's as_str but not `as_int`, `as_float`. Would be a nice thing to add

Issue #433, opened 2019-07-11
Support Integer/Float Min/Max Operations


naftulikay:
A feature that would be super helpful would be to implement a `min` and `max` function for generic number types.

First stage:

 - unsigned integers: upgrade all unsigned integers to a `u64` or a `usize`
 - floats: upgrade all floats to a `f64`
 - non-numbers: return an `Err`

Second stage:

 - unsigned/signed integer comparison: pretty straightforward comparison
 - float/integer comparison: also fairly straightforward

This would probably be a useful addition to the 1.0.0 release. I'd like to take on the work, but my availability isn't great right now.

Keats:
That does seem a bit niche and is easy to implement in someone's project if they need it. I don't think it makes sense to add it to Tera itself

naftulikay:
I know that Tera is not Jinja, [but Jinja does support these operations](http://jinja.pocoo.org/docs/2.10/templates/#max).

The specific use-case that we had was rendering two different systemd unit drop-ins to isolate two processes by CPU:

Template 1:

```jinja
[Service]
CPUAffinity=0-{{ max(0, floor(num_cpus / 2) - 1) }} 
```

Template 2:

```jinja
[Service]
CPUAffinity={{ max(0, floor(num_cpus / 2)) }}-{{ max(0, num_cpus - 1) }}
```

There are edge-cases which this template does not catch, but the idea would be that service A would be pinned from CPUs 0-7 and service B to CPUs 8-15 on a 16-core machine.

I'm still happy to contribute `min`, `max`, and `floor` as either filters or functions to Tera to add this functionality if there's a desire for it.

Keats:
hmm If they exist in Jinja2 it's different. The main issue is that we can't really support variadic functions so we are limited to 2 arguments and a function wouldn't be able to get the min/max of an array.

naftulikay:
Yes, my usage above was pseudocode, it's likely that these functions would need to be filters like Jinja does it.

Keats:
Closing as part of the Tera v2 release. If this is still relevant in v2, please reopen — see the 
  migration guide: https://github.com/Keats/tera/blob/master/MIGRATION.md

PR #448, opened 2019-09-03
int filter implementation

tones111:
This change set implements the [int](https://jinja.palletsprojects.com/en/2.10.x/templates/#int) filter.


Keats:
I guess you meant float filter rather than int?

tones111:
This PR implements both.  After having implemented int I noticed jinja has a float filter and the implementation was very similar, so I added a commit for it on the branch.

docs/content/docs/_index.md
```diff
@@ -1086,6 +1086,12 @@ Example: `{{ sections | get(key="posts/content") }}`
 Split a string into an array of strings, separated by a pattern given.
 Example: `{{ path | split(pat="/") }}`
 
+#### int
+Converts a value into an integer.  The `default` argument can be used to specify the value to return on error, and the `base` argument can be used to specify how to interpret the number.  Bases of 2, 8, and 16 understand the prefix 0b, 0o, 0x, respectively.
```
Keats: I would prefer starting by not having the bases, it seems extremely niche and I can't even think of a usecase for it

Keats:
@tones111 ping if you have time

tones111:
Sorry, I got a bit discouraged.  I'm really not sure what the goals or development plan is for this library.  Initially it appeared maintaining jinja compatibility was important, but subsequent efforts indicate a resistance to improving compatibility.  Issues or feature requests are met by responses that the functionality is to niche for incorporation which makes it harder to find the motivation to contribute.

There are several application areas (ex: source code generation) where template engines are useful and can benefit from the functionality that jinja supports but aren't implemented here.  If you prefer users of your library to independently implement those features on their own, then so be it.  You're welcome to make or amend the functionality here if you're interested in it.  I had intended for this to be a stepping stone toward better numeric support (#433  #434) and it would be nice to implement several of the jinja builtin comparison tests.

I apologize again for the rant, but I'm very confused on the direction and desires of this project.  What's here looks to have a lot of utility and I feel filling some of these gaps would only help further adoption.

Keats:
Sorry :(

>  Initially it appeared maintaining jinja compatibility was important

It isn't, taking any of my templates in Jinja2/Django would never work in Tera/Rust so I don't think compatibility is that important. The existing filters/tests are only there to have some kind of common base and were pulled from Liquid/Jinja2/Django documentations. Even the same filter with the same params is not compatible with Jinja since Tera requires keyword arguments for calls.

Porting an existing filter is ok (such as the int/float etc) but since compatibility isn't a goal, each feature needs to be weighted independently. Converting to int/float as mentioned in the two issues has a clear usecase, but the bases argument here less so. Is anyone actually using it or is it left for backward compatibility? Adding arguments is backward compatible so I generally prefer to expose the minimum and if needed extend it later so I know it's actually used.

PR #450, opened 2019-09-19
Add support for specifying timezones for `date` filter

### LINKED ISSUE #157: Support date timezone
Currently, formatting a timestamp with the `date` filter prints the time on the UTC timezone.
It would be useful to be able to specify the desired timezone, or to default on the local one.

maxdeviant:
This PR adds support for specifying a timezone in [IANA](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) format (e.g., `America/New_York`, `Asia/Shanghai`) when using the `date` filter.

Resolves #157.

### Motivation

When working on my Zola blog I noticed that when using UTC timestamps (generated with [`toISOString`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString)) the dates were always rendered as UTC time rather than local time.

### Usage

```handlebars
<!-- Outputs "2019-09-18" -->
{{ "2019-09-19T01:48:44.581Z" | date(timezone="America/New_York" }}
```

### Questions

- Currently this is only implemented for RFC3339 dates. Do we want to support this for i64 timestamps as well? I don't think `YYYY-MM-DD` dates apply since they don't have a time component.

- This might pertain to Zola rather than Tera, but should we add a configuration option for a default timezone to use?

### DIFF
--- a/src/builtins/filters/common.rs
+++ b/src/builtins/filters/common.rs
@@ -5,6 +5,7 @@ use std::iter::FromIterator;
 use crate::errors::{Error, Result};
 #[cfg(feature = "builtins")]
 use chrono::{DateTime, FixedOffset, NaiveDate, NaiveDateTime, Utc};
+#[cfg(feature = "builtins")]
 use chrono_tz::Tz;
 use serde_json::value::{to_value, Value};
 use serde_json::{to_string, to_string_pretty};
--- a/src/builtins/filters/common.rs
+++ b/src/builtins/filters/common.rs
@@ -5,6 +5,7 @@ use std::iter::FromIterator;
 use crate::errors::{Error, Result};
 #[cfg(feature = "builtins")]
 use chrono::{DateTime, FixedOffset, NaiveDate, NaiveDateTime, Utc};
+use chrono_tz::Tz;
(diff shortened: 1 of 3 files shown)


maxdeviant:
@Keats I can update the documentation as well, but I wanted to get your feedback on the implementation before I go through the trouble of updating the docs.

Keats:
Looks good to me! As you mentioned, only an update to the docs is missing

src/builtins/filters/common.rs
```diff
@@ -5,6 +5,7 @@ use std::iter::FromIterator;
 use crate::errors::{Error, Result};
 #[cfg(feature = "builtins")]
 use chrono::{DateTime, FixedOffset, NaiveDate, NaiveDateTime, Utc};
+use chrono_tz::Tz;
```
Keats: you need to add another `#[cfg(feature = "builtins")]` above that line

maxdeviant: @Keats Ah, thanks! I was wondering what was causing CI to fail despite it building locally.

Keats:
Thanks!

PR #452, opened 2019-09-23
Avoid panic on dividing by 0

vandenoever:
Dividing by 0 gives a panic because it leads to an unwrap().

Dividing by zero gives an infinite value. This patch adds a check for that.


Keats:
This has been fixed for a long time in the v1 version, 0.11 is not going to be updated anymore.

PR #453, opened 2019-09-24
Small doc fixes

gdetrez:
Examples for the `group_by` filter were using `sort`.


Keats:
Thanks for that, can you do the PR on the `next` branch instead?

gdetrez:
I don't see a `next` branch.  I can choose between `master` or `v1`.  Maybe it's not public?

Keats:
Sorry, I meant `v1`!

Keats:
Thanks!

PR #456, opened 2019-09-28
PoC: Render can take &Context

CAD97:
Note that this does the minimal threading of using `Context` at the root layer required to get this to compile; it might be worth it to see where else `Context` can be used instead of a raw `Value`.

I made `Tera::render` take `impl Borrow<Context>` here so that it can continue to take `Context` by value to avoid having to change all of the tests to pass `&Context`. The function potentially should always take `&Context`, though, and the benchmarks definitely should be adjusted to pass `&context` to avoid the initial clone at least.

### DIFF
--- a/benches/big_context.rs
+++ b/benches/big_context.rs
@@ -73,9 +73,9 @@ fn bench_big_loop_big_object(b: &mut test::Bencher) {
     .unwrap();
     let mut context = Context::new();
     context.insert("objects", &objects);
-    let rendering = tera.render("big_loop.html", context.clone()).expect("Good render");
+    let rendering = tera.render("big_loop.html", &context).expect("Good render");
     assert_eq!(&rendering[..], "0123");
-    b.iter(|| tera.render("big_loop.html", context.clone()));
+    b.iter(|| tera.render("big_loop.html", &context));
 }
 
 #[bench]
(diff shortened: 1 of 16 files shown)


Keats:
That looks pretty good! There was a reason it was taking Context by value but I can't remember it and if all tests pass it can't be that important. If everything works with `&Context`, we can change to that.

Keats:
@CAD97 do you have time to update the PR to take `&Context`?

Completely unrelated but since I know you are a Pest committer, do you have any idea on where the fix for https://github.com/pest-parser/pest/issues/402 would be or if pest3 is coming soon?

CAD97:
@Keats `render` now always takes `&Context`.

Pest3 is a long ways off and both it and issue 402 are blocked on dragostis finding time to work on it. Unfortunately nobody else has the knowledge required to move forward on either right now except him.

Keats:
I'll fix the last bug, thanks for that!

Keats:
While I'm there, @CAD97 any other API tweaks/changes you would like to see before the v1 release?

CAD97:
I think all of the API changes I'd put forward are already handled. There are other improvements I'd like to see (one big one: machine check more examples to avoid them getting out of date), but those are all semver-compatible to fix.

PR #458, opened 2019-10-28
Made the Error struct Send + Sync

### LINKED ISSUE #442: tera::Error cannot be sent between threads safely

AxelMontini:
In order to be able to construct `anyhow::Error`s from `tera::Error`s, we must make this crate errors `Send + Sync`.

To do so it's only necessary to change 3 lines of code, but I'm not sure whether this is going to be a breaking change or not.

IMO this is a small but great addition that contributes to the usability of this crate.

Edit:
Fixes #442

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -46,7 +46,7 @@ pub enum ErrorKind {
 pub struct Error {
     /// Kind of error
     pub kind: ErrorKind,
-    source: Option<Box<dyn StdError>>,
+    source: Option<Box<dyn StdError + Sync + Send>>,
 }
 
 impl fmt::Display for Error {
@@ -78,7 +78,7 @@ impl fmt::Display for Error {
 
 impl StdError for Error {
     fn source(&self) -> Option<&(dyn StdError + 'static)> {
-        self.source.as_ref().map(|c| &**c)
+        self.source.as_ref().map(|c| &**c as &(dyn StdError + 'static))
     }
 }
 
@@ -128,7 +128,7 @@ impl Error {
     }
 
     /// Creates generic error with a source
-    pub fn chain(value: impl ToString, source: impl Into<Box<dyn StdError>>) -> Self {
+    pub fn chain(value: impl ToString, source: impl Into<Box<dyn StdError + Send + Sync>>) -> Self {
         Self { kind: ErrorKind::Msg(value.to_string()), source: Some(source.into()) }
     }
 
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -165,16 +165,8 @@ pub type Result<T> = ::std::result::Result<T, Error>;
 mod tests {
     #[test]
     fn test_error_is_send_and_sync() {
-        trait SendAndSyncTrait: Send + Sync {
-            fn very_useless_function();
-        }
+        fn test_send_sync<T: Send + Sync>() {}
 
-        impl SendAndSyncTrait for super::Error {
-            fn very_useless_function() {
-                println!("Yay!");
-            }
-        }
-
-        super::Error::very_useless_function();
+        test_send_sync::<super::Error>();
     }
 }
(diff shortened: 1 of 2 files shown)


Keats:
Thanks, can you add a test like https://stackoverflow.com/questions/41599921/how-to-define-a-struct-that-is-sendsync-and-is-holding-a-iowrite-member?rq=1 to ensure it's correct?

AxelMontini:
Now the problem is that the CI fails because a command containing `&&` is being run on windows. Gonna create a PR for that too.

tests/render_fails.rs
```diff
@@ -142,3 +142,18 @@ fn test_error_in_macro_location() {
         "Failed to render 'error-location/error_in_macro.html': error while rendering macro `macros::cause_error`"
     );
 }
+
+#[test]
+fn test_error_is_send_and_sync() {
```
Keats: sorry, can you move it to a new test module in src/errors.rs so it's co-located?

Keats:
For the Windows failure I'm confused, it passes on https://github.com/Keats/tera/pull/353/checks?check_run_id=267363501

AxelMontini:
> For the Windows failure I'm confused, it passes on https://github.com/Keats/tera/pull/353/checks?check_run_id=267363501

Probably Beta stuff. I'm not even sure where to check for changes in GitHub Actions, but probably something has been changed in the last ~8 days.

PR to fix that too: #459

src/errors.rs
```diff
@@ -160,3 +160,21 @@ impl From<serde_json::Error> for Error {
 }
 /// Convenient wrapper around std::Result.
 pub type Result<T> = ::std::result::Result<T, Error>;
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test_error_is_send_and_sync() {
```
CAD97: Typically this is done with `fn test_send_sync<T: Send + Sync>() test_send_sync::<Error>()`. Note that `fn very_useless_function` isn't actually needed here; the `impl SendSync for Error` will fail if the supertraits aren't supplied.

There's no way to make it fail at runtime rather than compile time though, unless doing a stupidly complex "faux specialization" trick which really isn't worth it; SendSync is a test you want to stop you immediately if you break it anyway.

AxelMontini: *Improved*

Keats:
Thanks!

PR #459, opened 2019-10-28
And (&&) operator for windows

AxelMontini:
GitHub Actions CI fails on window because of the command `cd examples/actix && cargo build`.

The fix is to always use the bash shell, even on windows (git bash is used).

Edit:
I'm not sure why it's failing right now. Probably some recent changes to GitHub Actions broke the CI.


Keats:
Thanks!

PR #461, opened 2019-11-07
Fixes #460

### LINKED ISSUE #460: Strings with accents are wrongly sorted
As stated here: https://zola.discourse.group/t/how-to-sort-taxonomies-taking-in-account-accentuated-letters

We badly sort the strings with accents. "écriture" is sorted after "zola" for instance.

Geobert:
Fixes https://github.com/Keats/tera/issues/460

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ chrono = {version = "0.4", optional = true}
 chrono-tz = {version = "0.5", optional = true}
 # used in truncate filter
 unic-segment = {version = "0.9", optional = true}
+deunicode = {version = "1.0.0", optional = true}
 
 [dev-dependencies]
 serde_derive = "1.0"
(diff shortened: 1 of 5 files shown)


Keats:
That looks good but I don't know what the param name should be, `deunicode` is very hard to understand :/
Can you update the docs as well to mention that argument?

Geobert:
I agree that the param's name is not the best…
What about `lexicographical`? 
I was waiting your greenlight on the PR before updating the doc, will do it with the change of param's name :)

Keats:
I did a bit of research right now and it seems this function is not going to work until you can pass a locale, see https://stackoverflow.com/questions/1097908/how-do-i-sort-unicode-strings-alphabetically-in-python

Unless there is something similar in Rust, I'm afraid this isn't going to be merged since it would be incorrect for some languages.

Geobert:
I know Rust lack of collation support right now. But if this can't get merged I won't be able to use Zola…

Keats:
Changing it later would be a breaking change though and anyone writing in Swedish or other language would rightly report that as a bug :/ I don't see a good way to solve that until Rust gets some collation crate

Geobert:
Fair enough, feel free to close it :)

Keats:
I'm closing the PR but i'll update the issue with the details and leave that open

PR #466, opened 2019-11-26
Fix code examples in `group_by` docs

phil-opp:
They used the `sort` function instead of the `group_by` function that this section is about.


Keats:
Thanks! This is already fixed in the v1 branch but the website is still serving the old version docs. I should release Tera v1 very very soon anyway at some point it needs to happen :p

phil-opp:
Cool!

PR #468, opened 2019-12-05
Add #[derive(Clone)] to Tera struct

0x1793d1:
Hi,

I am using Tera v1 with actix-web v1. When constructing the app, I would prefer to clone Tera than parsing the templates as many times as there are workers.


Keats:
Thanks!

PR #473, opened 2019-12-11
Use usize instead of i64 in filesizeformat filter

0x1793d1:
Use of usize instead of i64 for 3 reasons:
- Become more permissive with upper values
- usize is more commonly used as file size unit
- humansize::file_size_opts::CONVENTIONAL doesn't allow negative values (and return an error in this case)


Keats:
Thanks!

PR #476, opened 2019-12-17
respect whitespace within filter section

nayato:
Prevent whitespace from getting trimmed within filter section.


Keats:
Thanks, I wish I remembered why it was that way :/

PR #477, opened 2019-12-18
Fix GlobalFn in docs

0x1793d1:
Wouldn't this be an opportunity to provide the alias GlobalFn in the API?


Keats:
Hmm the docs should be updated to rename GlobalFn -> Function as the name changed and I forgot to update the docs apparently.

agorton-nvl:
This issue remains outstanding.

Keats:
Fixed, thanks

Issue #480, opened 2020-01-04
maps | length


KoltesDigital:
Have you thought about `length` filter returning a map's length? Right now it only works for arrays and strings.

Ten0:
You could already add it manually thanks to the function import feature.

KoltesDigital:
I'm suggesting to make it built-in, given that both arrays and objects are both iterables using for loops, and that at least I wanted this filter multiple times on multiple Tera instances, I guess I'm not the only one it would benefit. The PR would have been just about adding one line, that's why I asked first if there's a reason. I can make this PR if you'd like.

Keats:
I think it makes sense to add, PR welcome!

KoltesDigital:
Here you go!

KoltesDigital:
BTW, how do you feel about length returning 0 for other types? Shouldn't an error be thrown instead?

Keats:
That would be a breaking change so only for the next major version.

Ten0:
> BTW, how do you feel about length returning 0 for other types? Shouldn't an error be thrown instead?

I agree! ^^
We're having a similar debate at https://github.com/Keats/tera/issues/481.

KoltesDigital:
Haha! At the same time.

I share your view of rust philosophy. Actually I see no use cases where length would be applied to a value not array|object|string, but I can imagine I would invert two variables by mistake. In this case, I'd prefer an error stating that I'm wrong, instead of silently working but not as expected.

Keats:
Actually for the `length` filter it should have been an error from the start for bool/numbers so changing that is a bugfix and is ok to do.

Keats:
It's in 1.0.2

PR #482, opened 2020-01-10
Add support for object in length built-in filter

KoltesDigital:
Discussed in #480.

### DIFF
--- a/docs/content/docs/_index.md
+++ b/docs/content/docs/_index.md
@@ -811,7 +811,7 @@ Example: `{{ value | join(sep=" // ") }}`
 If value is the array `['a', 'b', 'c']`, the output will be the string "a // b // c".
 
 #### length
-Returns the length of an array, an object, or a string. Always 0 for other types.
+Returns the length of an array, an object, or a string.
 
 #### reverse
 Returns a reversed string or array.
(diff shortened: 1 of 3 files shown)


docs/content/docs/_index.md
```diff
@@ -811,7 +811,7 @@ Example: `{{ value | join(sep=" // ") }}`
 If value is the array `['a', 'b', 'c']`, the output will be the string "a // b // c".
 
 #### length
-Returns the length of an array or a string, 0 if the value is not an array.
+Returns the length of an array, an object, or a string. Always 0 for other types.
```
Keats: Can you change the other types to error instead?

KoltesDigital: Done. The comments were removed, I found no statement about returning ("throwing") an error.

KoltesDigital:
BTW, I saw a lot of message (aka generic) errors which could be formalized as: filter applied on an unexpected value type, filter expects an argument which wasn't set, filter expects an argument with a certain type but was another type, etc. How do you feel about adding these variants in Error?

KoltesDigital:
Oops, too fast.

Keats:
I don't think there's much point in differentiating runtime error types, that would just create way too many error types (there are tests and fns that would get those too to be consistent) to be useful.

Keats:
Thanks for the PR!

PR #492, opened 2020-02-16
Add Tera::render_str

XAMPPRocky:
When creating a fluent integration with Tera, it is slightly annoying to need to register a template to be able to render a `Tera` instance with a custom function when testing. This PR adds `Tera::render_str` to handle that. It uses the code from `Tera::one_off`, and `Tera::one_off`, has been refactored to use this function instead. I opted to change the template name from `one_off` to `__one_off` in case anyone uses this for not testing it's less likely to conflict.

### DIFF
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -16,6 +16,9 @@ use crate::renderer::Renderer;
 use crate::template::Template;
 use crate::utils::escape_html;
 
+/// The of the the template used for `Tera::render_str` and `Tera::one_off`.
+const ONE_OFF_TEMPLATE_NAME: &str = "__tera_one_off";
+
 /// The escape function type definition
 pub type EscapeFn = fn(&str) -> String;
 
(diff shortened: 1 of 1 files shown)


src/tera.rs
```diff
@@ -306,11 +306,39 @@ impl Tera {
...
+    /// will add a template called `__one_off` to your instance. This method is
+    /// mainly useful for testing.
+    ///
+    /// Any errors will mention the `__one_off` template: this is the name given to the template by
+    /// Tera
+    ///
+    /// ```rust,ignore
+    /// let mut context = Context::new();
+    /// context.insert("greeting", &"hello");
+    /// Tera::one_off("{{ greeting }} world", &context, true);
```
Keats: it's not the right example

src/tera.rs
```diff
@@ -306,11 +306,39 @@ impl Tera {
...
+        input: &str,
+        context: &Context,
+        autoescape: bool,
+    ) -> Result<String> {
+        self.add_raw_template("__one_off", input)?;
+        if autoescape {
+            self.autoescape_on(vec!["__one_off"]);
+        }
+
+        self.render("__one_off", &context)
```
Keats: That function has the side effect of changing the auto-escape of the whole instance.
If that function is only really used for testing in practice, maybe we can just clone the Tera instance and change auto-espacing/adding the template on that new instance that gets dropped at the end of the function.

XAMPPRocky:
@Keats I've addressed those changes, and I've also changed it to use a more specific name and clean up after itself, so we shouldn't have to worry about it causing any side effects in peoples code.

src/tera.rs
```diff
@@ -306,11 +309,31 @@ impl Tera {
...
+    ///
+    /// ```rust,ignore
+    /// let mut context = Context::new();
+    /// context.insert("greeting", &"Hello");
+    /// let string = tera.render_str("{{ greeting }} World!", &context)?;
+    /// assert_eq!(string, "Hello World!");
+    /// ```
+    pub fn render_str(&mut self, input: &str, context: &Context) -> Result<String> {
+        self.add_raw_template(ONE_OFF_TEMPLATE_NAME, input)?;
+        let string = self.render(ONE_OFF_TEMPLATE_NAME, &context)?;
```
Keats: the template will still be there if the render fails, it needs to remove the template when it errors as well.

XAMPPRocky: Updated so now that it is always removed.

Keats:
Thanks it looks good. Sorry for the delay, I was moving countries.
I'll merge + deploy this weekend.

Keats:
Released in 1.1.0

PR #495, opened 2020-03-24
feat(filter): add more trim functions

guangie88:
Add additional whitespace trimming `trim_start` and `trim_end`.

Add `trim_start_matches` and `trim_end_matches` implementations.

### DIFF
--- a/docs/content/docs/_index.md
+++ b/docs/content/docs/_index.md
@@ -766,6 +766,26 @@ If value is "foo  bar", the output will be "Foo  Bar".
 #### trim
 Remove leading and trailing whitespace if the variable is a string.
 
+#### trim_start
+Remove leading whitespace if the variable is a string.
+
+#### trim_end
+Remove trailing whitespace if the variable is a string.
+
+#### trim_start_matches
+Remove leading characters that match the given pattern if the variable is a string.
+
+Example: `{{ value | trim_start_matches(pat="//") }}`
+
+If value is "//a/b/c//", the output will be "a/b/c//".
+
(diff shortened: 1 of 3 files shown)


guangie88:
Sorry that I didn't add any issue with regards to the PR, not sure what it is the convention to get a PR started.

Basically the issue that I was encountering was needing to trim away possible trailing forward slash for URL template value. While currently it is possible to do so, it requires pretty lengthy if-check `ending_with` to cater to two possible cases.

E.g.

```jinja
{% if SOME_URL is ending_with('/') %}{{ SOME_URL }}{% else %}{{ SOME_URL }}/{% endif %}data/xxx.json
```

can become

```jinja
{{ SOME_URL | trim_end_matches(pat="/") }}/data/xxx.json
```

Since Rust already has `trim_start_matches` (and the corresponding `trim_end_matches`) for `str`, I followed the current Tera `split` implementation and testing code for the above.

Additionally, since it is pretty weird to have `trim_start_matches` without `trim_start`, I trivially added in `trim_start` (and `trim_end`) to complement the current `trim` implementation.

Hope the feature can be added in if you think it is worthwhile (and hopefully we can get a v1.2.0 too :smiley:).

Keats:
The PR looks good and the additions are completely fine. Thanks for that!  I'll merge & release during the weekend.

guangie88:
> The PR looks good and the additions are completely fine. Thanks for that! I'll merge & release during the weekend.

Thank you for the response and comments, and I shall look forward to the new release! :+1:

PR #504, opened 2020-05-07
Add support for filters on array literals

azerupi:
Hi,
I went ahead and implemented support for filters on array literals even though I wasn't sure if this was considered a bug or not. :slightly_smiling_face: 

I am not sure if I covered every case? I made sure to add some tests on set blocks, for loops and kwargs. I also compiled zola with this patch and made sure that the code snippet [posted on the forum](https://zola.discourse.group/t/can-the-page-title-be-added-to-the-toc/404) was correctly parsed and did the same as the multi-line one.

Let me know if you are interested in this patch and if you want me to change some things or add some more tests.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -519,6 +519,7 @@ fn parse_variable_tag(pair: Pair<Rule>) -> TeraResult<Node> {
                 ws.right = p.as_span().as_str() == "-}}";
             }
             Rule::logic_expr => expr = Some(parse_logic_expr(p)?),
+            Rule::array_filter => expr = Some(parse_array_with_filters(p)?),
             _ => unreachable!("unexpected {:?} rule in parse_variable_tag", p.as_rule()),
         }
     }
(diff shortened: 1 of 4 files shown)


Keats:
That looks good I think. It's just missing a few tests in `src/renderer/tests` to easily test many cases.

azerupi:
I added two more tests in the renderer as you suggested. Is this enough?

Also when adding these tests I wondered if the construct below should be supported?

```tera
{{ [1, 2, 3] | length }}
```

I didn't add support for it because arrays were not supported before, so I didn't want to change it. But it feels like a logical extension.

Keats:
I don't remember any specific reason for that not existing, if it's not too hard it would be nice to support it.

azerupi:
No problem, it was easy to add.
If the number of tests is satisfactory, I think this PR is ready for a final review. :slightly_smiling_face:

Keats:
It looks good thanks! I'll merge/release later this week.

Issue #513, opened 2020-05-25
raw tag deletes whitespaces.


fx-kirin:
```
#[test]
fn make_sure_not_to_deleting_whitespaces() {
    let mut context = Context::new();
    context.insert("d", "d");
    let input = r#"{% raw %}    yaml_test:     {% endraw %}"#;
    let res = Tera::one_off(input, &context, true).unwrap();
    assert_eq!(res, "    yaml_test:    ");
}
```

I added the test like above. With the latest version `1.3.0`, all whitespaces after raw tag will be deleted. I think it should not. Is it appropriate workflow?

I tried to use this library with yaml file. But because of this, some workaround was needed to avoid this problem.

### Test result
```
running 1 test
thread 'renderer::tests::whitespace::make_sure_not_to_deleting_whitespaces' panicked at 'assertion failed: `(left == right)`
  left: `"yaml_test:"`,
 right: `"    yaml_test:    "`', src/renderer/tests/whitespace.rs:113:5
```

https://github.com/fx-kirin/tera/blob/eda23f6a1a74a300bccbc20ee95de801bd75e76a/src/renderer/tests/whitespace.rs#L107-L114

Keats:
Can you create a PR with that test?

Keats:
https://github.com/Keats/tera/pull/516 fixes it 
I'll check if there is anything I want to add to that release before publishing it.

PR #514, opened 2020-05-25
Added a test to check whitespaces deletion.

fx-kirin:
https://github.com/Keats/tera/issues/513


Keats:
Thanks, I'll have a look asap

PR #521, opened 2020-06-03
Add call context to any error in the processor

stanistan:
👋 

This is related to https://github.com/getzola/zola/pull/1041 - The patch attaches context to every `call` in the renderer/processor so that the name is in the "source" tree when its reported. So, `filter`, `function`, and `test` will each wrap an error produced from their invocation.


stanistan:
Thanks for merging, I haven't had a chance yet to get to adding tests for the rest, I can try to get to that in a follow up PR.

Keats:
No worries, I'm doing it. Thanks for the PR!

PR #516, opened 2020-05-27
Fix https://github.com/Keats/tera/issues/513

Keats:

PR #523, opened 2020-06-04
Make batch template sources generic

avitex:
- Backwards compatible
- Friendly API for less allocations on the consumer side

Cheers!

### DIFF
--- a/src/renderer/tests/whitespace.rs
+++ b/src/renderer/tests/whitespace.rs
@@ -103,7 +103,6 @@ fn works_with_filter_section() {
     assert_eq!(res, "  C   D");
 }
 
-
 #[test]
 fn make_sure_not_to_deleting_whitespaces() {
     let mut context = Context::new();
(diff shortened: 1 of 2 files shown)


avitex:
CI failing, see #516

Keats:
Can you add tests for example things you have in mind?

avitex:
Not sure what value that adds. You've already got test coverage for `vec![]`. Would just be testing other `IntoIterator`s

Keats:
I guess I was more curious about the usecase in practice.

avitex:
That's fair enough. An example use-case would be building a list of templates to be used,
as shown here. https://github.com/avitex/mdbook-tera/blob/master/src/lib.rs#L61

Rather than having to `collect` and allocate a vector, I would just pass the iterator in :)

Keats:
Good example, I'll merge that along the other PRs around the end of the week!

Issue #526, opened 2020-06-10
[Feature request] slice as python with negative index values


pando85:
Hello,

First of all, thank you for this project, it is amazing. Second I have a feature that could be interesting for this project.

I find very useful to use negative index values in slices. What do you think?

PR #527, opened 2020-06-10
Add feature negative slices indexes

pando85:
Resolves: #526


Keats:
Can you add a blurb on the docs mentioning that?

pando85:
@Keats sorry for being so late but I was on holidays. Done

PR #531, opened 2020-06-17
config.toml: Update netlify URL

j1mc:
Netlify has changed their default application URL from a netlify.com domain to netlify.app domain.
Updating the URL here will allow CSS to render correctly on the Tera docs website.


Keats:
Ah, thanks!

PR #537, opened 2020-07-23
feat(filter): add camel and snake filters

SirWindfield:
This PR adds two more filters for the common text transformations CamelText and snake_text. The implementation relies on `heck`, a text transformation library.

### DIFF
--- a/src/builtins/filters/string.rs
+++ b/src/builtins/filters/string.rs
@@ -418,6 +418,7 @@ mod tests {
         );
     }
 
+    #[cfg(feature = "builtins")]
     #[test]
     fn test_snake() {
         let result = snake(&to_value("Test Value").unwrap(), &HashMap::new());
@@ -425,6 +426,7 @@ mod tests {
         assert_eq!(result.unwrap(), to_value("test_value").unwrap());
     }
 
+    #[cfg(feature = "builtins")]
     #[test]
     fn test_camel() {
         let result = camel(&to_value("test value").unwrap(), &HashMap::new());
--- a/src/builtins/filters/string.rs
+++ b/src/builtins/filters/string.rs
@@ -80,7 +80,7 @@ pub fn lower(value: &Value, _: &HashMap<String, Value>) -> Result<Value> {
 }
 
 #[cfg(feature = "builtins")]
-pub fn snake(value: &Value, _: &HashMap<String, Value>) -> Result<Value> {
+pub fn snake_case(value: &Value, _: &HashMap<String, Value>) -> Result<Value> {
     use heck::SnakeCase;
     let s = try_get_value!("snake_case", "value", String, value);
 
@@ -88,7 +88,7 @@ pub fn snake(value: &Value, _: &HashMap<String, Value>) -> Result<Value> {
 }
 
 #[cfg(feature = "builtins")]
-pub fn camel(value: &Value, _: &HashMap<String, Value>) -> Result<Value> {
+pub fn camel_case(value: &Value, _: &HashMap<String, Value>) -> Result<Value> {
     use heck::CamelCase;
     let s = try_get_value!("snake_case", "value", String, value);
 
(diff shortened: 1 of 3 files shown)


SirWindfield:
On that note, there are a lot more transformations available, not sure if you want to add them to tera:
https://docs.rs/heck/0.3.1/heck/#cases-contained-in-this-library

I would also know your opinion on adding an alias for  `slugify` named `kebab`, as this is what a lot of people know the transformation under. A simple call to `tera.register_filter("kebab", string::slugify)` would be enough.

I hope I got the whole feature gating correct, if not, let me know!

SirWindfield:
No clue why the CI fails to be honest, running `cargo test` on my machine passes.

src/builtins/filters/string.rs
```diff
@@ -79,6 +79,22 @@ pub fn lower(value: &Value, _: &HashMap<String, Value>) -> Result<Value> {
     Ok(to_value(&s.to_lowercase()).unwrap())
 }
 
+#[cfg(feature = "builtins")]
+pub fn snake(value: &Value, _: &HashMap<String, Value>) -> Result<Value> {
```
Keats, line 83: can you rename all of it to snake_case and camel_case rather than snake/camel?

src/builtins/filters/string.rs
```diff
@@ -79,6 +79,22 @@ pub fn lower(value: &Value, _: &HashMap<String, Value>) -> Result<Value> {
...
+    use heck::SnakeCase;
+    let s = try_get_value!("snake_case", "value", String, value);
+
+    Ok(to_value(&s.to_snake_case()).unwrap())
+}
+
+#[cfg(feature = "builtins")]
+pub fn camel(value: &Value, _: &HashMap<String, Value>) -> Result<Value> {
+    use heck::CamelCase;
+    let s = try_get_value!("snake_case", "value", String, value);
```
Keats, line 93: camel_case

src/builtins/filters/string.rs
```diff
@@ -402,6 +418,20 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_snake() {
```
Keats, line 422: you need to feature gate those tests too

src/tera.rs
```diff
@@ -515,6 +515,8 @@ impl Tera {
     fn register_tera_filters(&mut self) {
         self.register_filter("upper", string::upper);
         self.register_filter("lower", string::lower);
+        self.register_filter("snake", string::snake);
```
Keats, line 518: as well as the lines adding the filters

SirWindfield:
> Heck doesn't seem maintained, is there an alternative? Something like https://github.com/serde-rs/serde/blob/master/serde_derive/src/internals/case.rs if it is on crates.io would be simpler and works for source code

I do get that, but https://lib.rs/crates/slug hasn't received an update either. `heck` was last updated after `slug`. And both repos seem quite the last two years.

Would you be OK with merging the PR using `heck` and I will create a new PR that removes both dependencies and re-uses the `serde` code, removing two dependencies at once?

SirWindfield:
On that note, I opened an issue on the repository to clarify the maintenance status and offered my help on there. Let's see how this turns out! :)

Keats:
slugify is kind of a "done" thing, no need for new releases. There are a few issues on heck repo though and I'm not sure casing needs to happen on unicode characters - that seems a bit overkill for code casing.

SirWindfield:
I mean the crate hasn't received any new love for 2 years. I'd just assume the issues as done at this point. From my side, it's the same argument. heck is a done thing, casing doesn't change.

You have the decision power though :) The implementations are rather easy so including it into tera might be a good idea. However, that only works for ASCII as unicode isn't taken into account and has to be documented IMO.

Keats:
In the end I think it's better to not add additional filters that pull dependencies and those are very easy to add if needed. The builtins feature already pulls a lot of dependencies.

SirWindfield:
I've created a crate that contains filters for Tera: https://github.com/SirWindfield/tera-text-filters
If anyone stumbles across this :)

PR #540, opened 2020-07-24
feat(context): add `contains_key` and `get` methods

### LINKED ISSUE #538: Add methods for checking if a value is inside a `Context`
I have a use-case where my application and context creation is split into multiple modules. I have to check if specific values are already inside the context and only insert if they are not. Currently there is no such method available. I'd propose these ones:

```rust
fn contains<S: Into<String>>(&self, key: S) -> bool;
fn get<S: Into<String>>(&self, key: S) -> Option<&Value>;
```

This can only be done currently by creating a custom type and implementing `From<type> for Context {}`

SirWindfield:
Closes #538

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -81,8 +81,14 @@ impl Context {
         Context::from_value(obj)
     }
 
-    pub(crate) fn get(&self, index: &str) -> Option<&Value> {
+    /// Returns the value at a given key index.
+    pub fn get(&self, index: &str) -> Option<&Value> {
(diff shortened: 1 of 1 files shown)


Keats:
Thanks!

SirWindfield:
Any plans when this gets released? It's not urgent. But if you plan to do so the next couple of days I'll just wait instead of fiddling around with my workaround :)

Keats:
1.4.0 is up

PR #545, opened 2020-08-09
Enabled filters & functions to be safe

Kixiron:
Allow external Rust functions and filters to be unescaped. Needed for [docs.rs](https://github.com/Kixiron/docs.rs/blob/features/templates/rustdoc/body.html#L79-L82) and for making more complex rust plugins without having to call `safe` within templates

### DIFF
--- a/src/builtins/filters/mod.rs
+++ b/src/builtins/filters/mod.rs
@@ -13,6 +13,11 @@ pub mod string;
 pub trait Filter: Sync + Send {
     /// The filter function type definition
     fn filter(&self, value: &Value, args: &HashMap<String, Value>) -> Result<Value>;
+
+    /// Whether the current filter's output should be treated as safe, defaults to `false`
+    fn is_safe(&self) -> bool {
+        false
+    }
 }
 
 impl<F> Filter for F
--- a/src/builtins/functions.rs
+++ b/src/builtins/functions.rs
@@ -12,6 +12,11 @@ use crate::errors::{Error, Result};
 pub trait Function: Sync + Send {
     /// The global function type definition
     fn call(&self, args: &HashMap<String, Value>) -> Result<Value>;
+
+    /// Whether the current function's output should be treated as safe, defaults to `false`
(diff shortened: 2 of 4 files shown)


Keats:
Very nice solution! I'll merge + release a new version later today or tomorrow
Also TIL docs.rs uses Tera!

Kixiron:
We do, Tera is a fantastic template engine and *miles* better than our [previous](https://github.com/rust-lang/docs.rs/blob/9883f210232df2e7ecf0f0bfeae72c4437a3ddc9/templates/crate_details.hbs)

d4h0:
@Kixiron: Thank you! I just was looking for a way to archive this.

@Keats: Was this already published with the last update? If so: it seems this new feature [isn't documented at the moment](https://tera.netlify.app/docs/#functions).

pietroalbini:
It's documented in [the API reference](https://docs.rs/tera/1.5.0/tera/trait.Function.html#method.is_safe).

d4h0:
@pietroalbini: Yes, I saw this before I found this PR. But I think for many Rust beginners (like myself) this will not be enough. In 6 months of using Rust I never had to create a closure like it is needed here, and only after I saw [safe_function_works](https://github.com/Keats/tera/blob/58c28cf7396adb84d007ae8c5f7fc2bff360d4aa/src/renderer/tests/basic.rs#L923), I realized how to use this new feature.

Therefore it would be useful for beginners if there was an example.

I was also assuming that https://tera.netlify.app/docs/ contains all important API documentation.

Keats:
https://tera.netlify.app/docs/ contains _some_ API information about common usecases (like loading and rendering a template). The function bit is actually outdated. I don't want to repeat the whole Rust API docs in it but I should at least put links to the relevant items in the rustdoc.

d4h0:
@Keats: Alright, thanks for the clarification. I assumed this feature is for a common use-case (if someone wants to use custom functions, at least).

PR #548, opened 2020-08-14
Macro namespace tests for issue 532

sooda:
All of these are expected to pass, but at the moment they just demonstrate a bug:

running 6 tests
test test_extend_and_include4 ... ok
test test_extend_and_include1 ... ok
test test_extend_and_include2 ... ok
test test_extend_and_include_nested2 ... ok
test test_extend_and_include3 ... FAILED
test test_extend_and_include_nested1 ... FAILED

See the comments in the template files for experiments.

### DIFF
--- a/tests/macros.rs
+++ b/tests/macros.rs
@@ -43,3 +43,21 @@ fn test_extend_and_include4() {
     assert_eq!(result.is_err(), false);
     assert_eq!(result.unwrap(), "hello");
 }
+
+#[test]
+fn test_extend_and_include_nested1() {
+    // inheritance, a macro import, another nested import, nested uses self
+    let result = render_tpl("extend_and_include_nested1.html");
+
+    assert_eq!(result.is_err(), false);
+    assert_eq!(result.unwrap(), "hello");
+}
+
+#[test]
+fn test_extend_and_include_nested2() {
+    // like test_extend_and_include_nested1 but no inheritance
+    let result = render_tpl("extend_and_include_nested2.html");
+
+    assert_eq!(result.is_err(), false);
+    assert_eq!(result.unwrap(), "hello");
+}
--- a/tests/macros.rs
+++ b/tests/macros.rs
@@ -1,0 +1,45 @@
+extern crate tera;
+
+use tera::{Context, Result, Tera};
+
+fn render_tpl(tpl_name: &str) -> Result<String> {
+    let tera = Tera::new("tests/macros/**/*.html").unwrap();
+
+    tera.render(tpl_name, &Context::new())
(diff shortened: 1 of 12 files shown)


tests/macros/macro_lib_a.html
```diff
@@ -0,0 +1,5 @@
+{% macro inner_impl() %}hello{% endmacro inner_impl -%}
+{# https://github.com/Keats/tera/issues/532
+   Try changing self:: to lib_a:: to pass one test and fail another.
```
sooda, line 3: That way we'd see the lib_a import in the test number 3, but the test number 4 would start failing because the imported macro doesn't have lib_a visible there, only self.

tests/macros/macro_lib_a.html
```diff
@@ -0,0 +1,5 @@
...
+   Try changing ::inner_impl() to ::unused() to fail in another way. #}
```
sooda, line 4: Here we'd see the content from lib_b. The self namespace does exist, but it appears to be assigned to the first import of the two in the rendered template.

tests/macros/macro_lib_d.html
```diff
@@ -0,0 +1,4 @@
+{% macro inner_impl() %}hello{% endmacro inner_impl -%}
+{# https://github.com/Keats/tera/issues/532
+   Try changing self:: to lib_d:: to pass one test and fail another. #}
```
sooda, line 3: Like in the first commit, this would get the first test to pass but not the second that uses no inheritance.

tests/macros.rs
```diff
@@ -0,0 +1,45 @@
...
+    let result = render_tpl("extend_and_include3.html");
+
+    assert_eq!(result.is_err(), false);
+    assert_eq!(result.unwrap(), "hello");
+}
+
+#[test]
+fn test_extend_and_include4() {
+    // like test_extend_and_include3 but no inheritance for completeness
+    let result = render_tpl("extend_and_include4.html");
```
sooda, line 41: The name ("extend_...") is a bit odd because the file doesn't actually extend anything, but it's related to the "extend from stuff and include macros" test cases, so I kept them all consistent.

johndoudou:
self:: does not seem to work in macros on my side.
I copied the example of the doc and it complains that :
`Failed to render 'section.html': error while rendering macro `mydebug::factorial` Reason: Macro `self::factorial` not found in template `section.html``

section.html uses an `{% extends "base.html" %}` so it may give some credit to the problem you mentioned, `The self namespace does exist, but it appears to be assigned to the first import of the two in the rendered template.`

PR #549, opened 2020-08-16
Allow multiline function kwargs with trailing comma

joshleeb:
This patch extends the `kwargs` pest grammar to allow multiline
functions with an optional trailing comma.


Keats:
Thanks!

PR #568, opened 2020-11-06
escape_html: use literal forward slash

### LINKED ISSUE #567: Invalid HTML entity encoding
If you use something like one of these:

~~~
<a href="{{ config.base_url }}">March</a>
<a href="{{ get_url(path="/") }}">March</a>
~~~

You get these results with `zola serve` and `zola build`, respectively:

~~~
<a href="http:&#x2F;&#x2F;127.0.0.1:1111">March</a>
<a href="&#x2F;winter">March</a>
~~~

This isnt a valid application of HTML entity encoding. The resultant links end
up being the same as far as the browser is concerned, but it just an unneeded
usage of the encoding. The only characters that should be encoded would be the
reserved characters `&`, `<`, `>` and `"`:

https://developer.mozilla.org/Glossary/Entity

a8nJ:
Forward slash is not a reserved HTML character, and most major implementations
do not encode it.

Fixes #567

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -2,9 +2,7 @@
 ///
 /// Escape the following characters with HTML entity encoding to prevent switching
 /// into any execution context, such as script, style, or event handlers. Using
-/// hex entities is recommended in the spec. In addition to the 5 characters
-/// significant in XML (&, <, >, ", '), the forward slash is included as it helps
-/// to end an HTML entity.
+/// hex entities is recommended in the spec.
 ///
 /// ```text
 /// & --> &amp;
@@ -12,7 +10,6 @@
 /// > --> &gt;
 /// " --> &quot;
 /// ' --> &#x27;     &apos; is not recommended
-/// / --> &#x2F;     forward slash is included as it helps end an HTML entity
 /// ```
 #[inline]
 pub fn escape_html(input: &str) -> String {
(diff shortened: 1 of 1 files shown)


Keats:
This is following OWASP recommendations (https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#rule-1-html-encode-before-inserting-untrusted-data-into-html-element-content) and I don't think escaping it breaks anything

a8nJ:
I mean suit yourself, but just check around the web and see what sites are (or are not) encoding forward slashes. For example, github.com:

    <link rel="dns-prefetch" href="https://github.githubassets.com">

google.com:

    <html itemscope="" itemtype="http://schema.org/WebPage" lang="en">

stackoverflow.com:

    <meta property="og:url" content="https://stackoverflow.com/"/>

reddit.com:

    <link rel="canonical" href="https://www.reddit.com/" />

even the OWASP site itself isnt doing it:

    <link rel="canonical" href="https://owasp.org/" />

a8nJ:
@Keats FYI, forward slash no longer exists in that link you gave.

I did a Pull Request to OWASP over the weekend.

Keats:
Thanks, I'll merge it for the next major version

89z:
https://github.com/Keats/tera/pull/662

PR #578, opened 2020-11-30
Fix extra memory allocation

jstelzer:
I'm new to rust and could be wrong, but I don't think you need to allocate this much memory.
push_str allocates the 'escaped' characters?

Let me start off by saying I'm only just learning rust. I've been doing what everyone does and wrote a toy blog framework using a bunch of crates and a lot of docs.

This got me into tera. One of my blog posts had deliberately malformed html tags in it just to see what would happen. I was delighted when it was magically escaped. So I went on a hunt to see where the magic lived.

Here we are.

When I read the utility function I kind of scratched my head because you were allocating len * 2 and then potentially requiring much more space than that depending on how many escaped chars there are.

So, I leveraged your test cases, added a test and it failed to fail. Then I figured out how to use your bench tools on rust nightly and they all passed with my change. It seems push_str is doing the memory allocation for us?

So, here's a PR. This is totally a silly case of me being curious. But if my change is indeed bad or would break something, please teach me. As I said, I only just started learning rust.

### DIFF
--- a/src/utils.rs
+++ b/src/utils.rs
@@ -16,7 +16,7 @@
 /// ```
 #[inline]
 pub fn escape_html(input: &str) -> String {
-    let mut output = String::with_capacity(input.len() * 2);
+    let mut output = String::with_capacity(input.len());
     for c in input.chars() {
         match c {
             '&' => output.push_str("&amp;"),
@@ -41,6 +41,7 @@ mod tests {
     fn test_escape_html() {
         let tests = vec![
             (r"", ""),
+            (r"&&&&&&&&", "&amp;&amp;&amp;&amp;&amp;&amp;&amp;&amp;"),
             (r"a&b", "a&amp;b"),
             (r"<a", "&lt;a"),
             (r">a", "&gt;a"),


Keats:
I don't remember why it is doing that.
An approach like https://github.com/SergioBenitez/Rocket/blob/master/core/http/src/raw_str.rs#L250 would be better but returning Cow would be a breaking change..
We can have the same approach without Cow and for the next major version we switch it.

genusistimelord:
Sorry if this is a really late reply to this. 

Generally, most programmers would double the size at the start or do this within a Power of 2 to avoid multiple Re-allocations when the data does not fit within the String or Array. As it is much slower to constantly Resize an Array, so the aspect is to do less resizing by setting the size higher. This is a double edge sword as it focuses on the aspect that it is fine to use more memory to increase the speed at which an array push is handled. It does heavily speed up the pushes but only if the data being inserted in was more than what the original size would have been, hence avoiding a resize of the array. Otherwise, you end up wasting ram and gaining no benefit if the data fit within the original Size.

The reason array allocation generally is slow is due to the OS finding a spot on the Ram to place it. It then will Allocate the new Spot at the new location, but also needs to be able to fit the Array within the spot with no fragmentation. If it cannot then it must find another spot till it can fit or throw a memory size error. I do not remember if the OS will allow Fragmentation or not but in most case's I would say this is not allowed for an Array.

I would probably investigate how often you see the array gets resized when in its original length as I thought Rust automatically Resized them larger than what it should be to reduce the number of times it needed to be resized during growth. Also since this is coming from a &str then its getting the Exact length which is then being expanded into the HTML correct sequences. so *2 might be a good start since some character expand to 3 to 5 characters long.

Issue #265, opened 2018-03-01
Allow block to be used multiple times


faulesocke:
It should be possible to use a block multiple times in the template. This is possible in jinja2, because a block can be accessed through the special `self` variable as described for example here:

https://stackoverflow.com/questions/20929241/how-to-repeat-a-block-in-a-jinja2-template

Keats:
I have no strong feeling about that. I don't really see the usecase though so I'd like to see more people wanting that

faulesocke:
Probably the most common use case is for the page title as seen in the stackoverflow thread. Of course one can easily work around this, but I thought it would be nice, if I (and others) don't have to.

paulcmal:
> Probably the most common use case is for the page title as seen in the stackoverflow thread.

Yes, that's a very common use-case. I also had to scratch my head around it when making a Hugo theme. I just don't understand the logic of not being able to reuse a block elsewhere in the template ^^" (is there an actual reason?)

Keats:
>  I just don't understand the logic of not being able to reuse a block elsewhere in the template ^^" (is there an actual reason?)

In the first template of the stackoverflow, the`title` block is defined twice. When using the `title` block in a child template, which one should the template engine inherits from? The template engine has no way of knowing which one so that's the reason you go through `self` in Jinja2

paulcmal:
> that's the reason you go through self in Jinja2

Thanks for the explanation! Am i wrong to think this could also be addressed differently by using a separate function for overriding the block? (such as [define in Hugo](https://gohugo.io/templates/base/#override-the-base-template))

Keats:
Define in Hugo is just block in Tera. What you need is a way to render the
content of a block, not redefine it. I'm not sure if the golang template
engine has a way to do that

paulcmal:
> What you need is a way to render the content of a block, not redefine it.

Yes, I meant that we need different keywords for defining and using a block. Sorry if i wasn't clear. I understand that it would make sense in order to avoid a breaking change to have define a `{% useblock BLOCK %}` feature. Would that be hard to implement?

I think the title example from stackoverflow is a clear example of why people would want this. Are variables the appropriate way to deal with such issues? (I'm not sure about the flow of data of a template to/from other templates it extends)

Keats:
Woops didn't see @paulcmal message

I think `useblock` makes sense and shouldn't be too hard to implement

BertalanD:
What's wrong with the `{{ self.block_name() }}` syntax?

[...]

PR #581, opened 2020-12-04
Implement `useblock` tag to expand blocks multiple times

christoph-heiss:
Hi!

So after using Tera I got to the same problem as discussed in #265, which is to expand a block multiple times. Since a new [`useblock` tag](https://github.com/Keats/tera/issues/265#issuecomment-442858181) was the proposed solution I though I'd implement that. The acutal core changes (136e1e1687fc665163c41cd299387190d2b714cd) aren't very invasive IMHO.

All test cases succeed, so existing users shouldn't be affected in any way due to being a completely new tag.

I added relevant test cases and updated the documentation where appriopriate. I hope that is ok as-is, but I'd be happy to revise it.

### DIFF
--- a/src/renderer/tests/basic.rs
+++ b/src/renderer/tests/basic.rs
@@ -571,6 +571,82 @@ fn ok_many_variable_blocks() {
 }
 
 #[test]
+fn can_expand_block_multiple_times_with_useblock() {
+    let result = render_template(
+        r#"
+{%- block test -%}Hello World{%- endblock test -%}
+{%- useblock test %} {% useblock test -%}
+        "#,
+        &Context::new()
+    );
+
+    assert_eq!(result.unwrap(), "Hello World Hello World");
+}
+
+#[test]
+fn can_expand_block_multiple_times_with_useblock_nested() {
+    let result = render_template(
+        r#"
+{%- block test -%}Hello World{%- endblock test -%}
+{%- block test2 -%}{%- useblock test %} {% useblock test -%}{%- endblock test2 -%}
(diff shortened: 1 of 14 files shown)


src/renderer/tests/basic.rs
```diff
@@ -570,6 +570,82 @@ fn ok_many_variable_blocks() {
...
+    assert_eq!(result.unwrap(), "Hello World Hello World");
+}
+
+#[test]
+fn can_expand_block_with_useblock_with_inheritance() {
+    let mut tera = Tera::default();
+    tera.add_raw_template("base", "{% useblock test %}").unwrap();
+    tera.add_raw_template(
+        "index",
+        "{% extends \"base\" %}{% block test %}Hello World{% endblock test %}"
```
Keats, line 606: that seems a bit wrong to me, this is not re-using the block test it's declaring it first in the base template via useblock and then overriding it in the child template? This shouldn't work

Keats (changes requested):

src/renderer/tests/basic.rs
jk-ozlabs: I think the intention here (unless I'm missing something) is to have the content of the block defined by the upper-level template, and referenced in the base.

The example from #265 was a title block (with content provided from the "index" template), which gets used within both a `<title>` tag and a `<meta>` tag in the base template.

jk-ozlabs:
@christoph-heiss are you still interested in this change? If not, I'm happy to continue your work here, if that's OK with you.

Issue #571, opened 2020-11-11
Add a more detailed example for GROUP_BY


rimutaka:
Issue #393 helped me figure out how to use GROUP_BY with this example:

```
{% set map = section.pages | group_by(attribute="year") %}
{% set_global years = [] %}
{% for year, ignored in map %}
  {% set_global years = years | concat(with=year) %}
{% endfor %}
{% for year in years | reverse %}
   {% set posts = map[year] %}
{% endfor %}
```
It may be worth adding it to the docs. @Keats , Vincent, will you consider a PR for this?

Keats:
Sure

Issue #393, opened 2019-03-01
accessing keys and values of a map


tcurdt:
I am trying to access the keys of hashmap and then lookup via keys. But I am a little lost when looking at the docs. This is what I tried:

```
{% set map = section.pages | group_by(attribute="year") %}
{% for year in map.keys() | reverse %}
  {% set posts = map[year] %}
```

Keats:
Tera templates cannot execute Rust code like `.keys()` and there is no way to iterate only on the keys of a hashmap, it could be an interesting improvement though.

Taking the context of your issue in Zola, something like that might work:

```jinja2
{% set map = section.pages | group_by(attribute="year") %}
{% set_global years = [] %}
{% for year, ignored in map %}
  {% set_global years = years | concat(with=year) %}
{% endfor %}
{% for year in years | reverse %}
   {% set posts = map[year] %}
{% endfor %}
```

Not great but we don't have the luxury of using Python like Jinja2/Django :/

tcurdt:
Wow, that's really really ugly :) but thanks for the work around!
I don't understand why the `global` is necessary though.
And yes - exposing the keys and values as arrays would be a nice addition then.

Keats:
> I don't understand why the `global` is necessary though

`set` is local, so calling `set` in a forloop only sets the value for the current iteration, which isn't what we want here. This is is to avoid having to use workarounds like https://stackoverflow.com/questions/9486393/jinja2-change-the-value-of-a-variable-inside-a-loop

tcurdt:
Oh, OK. Wasn't aware that jinja2 has such ugly corners, too. Thanks for the explanation.

PR #584, opened 2020-12-07
Added detailed example for GROUP_BY #571

rimutaka:
This PR combines an example from #393 with a slightly more detailed example on the basic [GROUP_BY](https://tera.netlify.app/docs/#group-by) usage.

Closes: #571

#### Background

I somehow missed the detail of the current GROUP_BY example and went on a tangent, that lead me to #393 with a good example on how to use GROUP_BY.

It takes a bit more space compared to other built-ins, but I think it will help others understanding how this feature can be used.

#### Caveat

I hope I got the syntax right. I had to modify it slightly from the actual code I used in my project, so this exact version was not tested.

### DIFF
--- a/docs/content/docs/_index.md
+++ b/docs/content/docs/_index.md
@@ -974,7 +974,27 @@ The `attribute` argument can be used to group posts by year:
 or by author name:
 
 ```jinja2
-{{ posts | group_by(attribute="author.name") }}
+{% for name, author_posts in posts | group_by(attribute="author.name") %}
+    {{ name }}
+    {% for post in author_posts %}
+        {{ post.year }}: {{ post.content }}
+    {% endfor %}
+{% endfor %}
+```
+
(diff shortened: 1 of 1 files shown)


rimutaka:
It looks like the checks fail for an unrelated reason.

docs/content/docs/_index.md
```diff
@@ -974,7 +974,27 @@ The `attribute` argument can be used to group posts by year:
...
+
+```jinja2
+{% set map = section.pages | group_by(attribute="year") %}
+{% set_global years = [] %}
+{% for year, ignored in map %}
+    {% set_global years = years | concat(with=year) %}
+{% endfor %}
+{% for year in years | reverse %}
+    {% set posts = map[year] %}
+{% endfor %}
```
Keats, line 997: that's pretty smart, I don't think I would have thought of that

PR #597, opened 2021-01-26
Enhance include tag with list and "ignore missing" support

richardchien:
Hi! I've added "list of templates" and "ignore missing" support for include tag, so that we can write include tags like below, which have the same syntax as Jinja2 (https://jinja.palletsprojects.com/en/2.11.x/templates/#include).

```jinja
{% include "sidebar.html" ignore missing %}
{% include ['page_detailed.html', 'page.html'] %}
{% include ['special_sidebar.html', 'sidebar.html'] ignore missing %}
```


Keats:
I didn't even know it existed but the PR looks ok. I'll merge it this weekend when I do a release

PR #601, opened 2021-02-05
Use std::io::Write to render templates

xabufr:
This allows users to render templates directly to anything that implements std::io::Write (for exemple files).

These changes permits users to control the output buffer when rendering to memory. They should also reduce number of memory allocations, as single rendering steps doesn't return Strings anymore (but instead uses output buffer).

Now the only case where memory really needs to be allocated is when we do not know when the rendered value should really but append to output.

### DIFF
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -8,10 +8,13 @@ mod macros;
 mod processor;
 mod stack_frame;
 
+use std::io::Write;
+
 use self::processor::Processor;
 use crate::errors::Result;
 use crate::template::Template;
 use crate::tera::Tera;
+use crate::utils::buffer_to_string;
 use crate::Context;
 
 /// Given a `Tera` and reference to `Template` and a `Context`, renders text
@@ -44,15 +47,16 @@ impl<'a> Renderer<'a> {
 
     /// Combines the context with the Template to generate the end result
     pub fn render(&self) -> Result<String> {
-        let output;
+        let mut output = Vec::with_capacity(10_000);
+        self.render_to(&mut output)?;
+        Ok(buffer_to_string(output))
+    }
 
-        {
-            let mut processor =
-                Processor::new(self.template, self.tera, &self.context, self.should_escape);
+    /// Combines the context with the Template to write the end result to output
+    pub fn render_to(&self, mut output: impl Write) -> Result<()> {
+        let mut processor =
+            Processor::new(self.template, self.tera, &self.context, self.should_escape);
 
-            output = processor.render()?;
-        }
-
-        Ok(output)
+        processor.render(&mut output)
     }
 }
--- a/src/renderer/mod.rs
+++ b/src/renderer/mod.rs
@@ -49,7 +49,7 @@ impl<'a> Renderer<'a> {
     pub fn render(&self) -> Result<String> {
         let mut output = Vec::with_capacity(10_000);
         self.render_to(&mut output)?;
-        Ok(buffer_to_string(output))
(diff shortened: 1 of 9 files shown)


src/renderer/mod.rs
```diff
@@ -44,15 +47,16 @@ impl<'a> Renderer<'a> {
 
     /// Combines the context with the Template to generate the end result
     pub fn render(&self) -> Result<String> {
-        let output;
-
-        {
-            let mut processor =
-                Processor::new(self.template, self.tera, &self.context, self.should_escape);
+        let mut output = Vec::with_capacity(10_000);
```
Keats, line 50: any reason 10k was chosen?

src/renderer/processor.rs
```diff
@@ -145,21 +147,19 @@ impl<'a> Processor<'a> {
         }
     }
 
-    fn render_body(&mut self, body: &'a [Node]) -> Result<String> {
-        let mut output = String::with_capacity(body.len() * 20);
-
+    fn render_body(&mut self, body: &'a [Node], w: &mut impl Write) -> Result<()> {
```
Keats, line 150: can we rename `w` -> `writer` everwhere?

src/utils.rs
```diff
@@ -33,9 +35,24 @@ pub fn escape_html(input: &str) -> String {
...
+    F: FnOnce(&mut Vec<u8>) -> Result<(), E>,
+    Error: From<E>,
+{
+    let mut buffer = Vec::new();
+    render(&mut buffer).map_err(Error::from)?;
+    Ok(buffer_to_string(buffer))
+}
+
+pub(crate) fn buffer_to_string(buffer: Vec<u8>) -> String {
+    String::from_utf8(buffer).expect("Render only UTF-8 bytes")
```
Keats, line 49: let's handle that error as well rather than `expect`

Keats (changes requested):
Looks good.
Can you also add a benchmark for `render_to` to see the difference?

src/renderer/mod.rs
xabufr: The only reason is that previous code allocate a String buffer of 10.000 chars: https://github.com/Keats/tera/pull/601/files#diff-e67fd4dbfd06da05000d6b37dcecbf48e765d6e30eb90d021fa01fae7b094887L1027

src/renderer/processor.rs
xabufr: Yes I will did it

src/utils.rs
xabufr: Okay, but this will introduce a new error type (and at the moment this error should never happen as we only write UTF8 bytes)

Keats: It's probably ok, we have non-exhaustive error enums.

src/renderer/mod.rs
Keats: oh I forgot about that

src/renderer/processor.rs
xabufr: done

src/utils.rs
xabufr: It should be ok now

Keats:
Thanks for the changes! Can you add a benchmark?

xabufr:
Yes I'll try to do that when I had time

Keats:
I'll check to add it myself, thanks!

freedeaths:
Excuse me, when will this "render_to" feature in "next" branch be released in a stable version?

Keats:
Not long, maybe this weekend? Probably a couple more bug fixes to go in https://github.com/Keats/tera/pull/607 if possible

PR #602, opened 2021-02-09
Follow symlinks in glob

max-frai:
In some complicated dev environments templates folder could be in other place, but located near executable via symlinks. This pull request enables follow links argument for globwalk builder.


Keats:
Thanks!

Issue #294, opened 2018-05-22
Add wasm target


Keats:
AFAIK only chrono fails to compile so disabling/tweaking all date related filters/global functions should do the trick

Keats:
So `cargo +nightly build --release --target wasm32-unknown-unknown` does build now.
The question is: what do I do now? Anyone has an idea?

Keats:
A good idea would be a playground on the documentation site where you can write some templates, set some data and render it

pickfire:
Is the idea of playground still valid? I wanted to try tera online but I can't seemed to find any playground yet to easily do that.

Keats:
The wasm target works but there is no playground right now. It would be great though.

PR #608, opened 2021-02-27
Add playground

pickfire:
I feel lazy to do stuff related to zola fix. Right now it needs some changes which I think we need to change the CI unless we included the built wasm here. We may also need to do some styling (which I am not so good at, lazy...).

To get it working, I can't seemed to get it working with zola due to how the stuff was generated in wasm-pack. Instructions (development, production just add a few more `--release`):

```
cd docs/tera-playground
wasm-pack build
cd ..
npm ci
npm run build
zola build
sd 'https:&#x2F;&#x2F;tera.netlify.app' 'http:&#x2F;&#x2F;127.0.0.1:8000' public/index.html
python -m http.server --directory public/
```

Test
```
cd docs/tera-playgound
wasm-pack test --firefox
```

Good

image

Bad

image

From https://github.com/Keats/tera/issues/294#issuecomment-786807725

Would be sad if I do CI and CSS but I want to hear out what you think first.

At least this could let me use tera on the web at least, no need to manually guess if it will produce whitespaces for these kind of issues like https://github.com/rust-lang/docs.rs/pull/1160 and https://github.com/rust-lang/docs.rs/pull/1288, can just try in tera playground easily.

### DIFF
--- a/docs/tera-playground/Cargo.toml
+++ b/docs/tera-playground/Cargo.toml
@@ -1,0 +1,37 @@
+[package]
+name = "tera-playground"
+version = "0.1.0"
+authors = ["Ivan Tham <pickfire@riseup.net>"]
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[features]
+default = ["console_error_panic_hook"]
+
+[dependencies]
+wasm-bindgen = "0.2.63"
+
+# The `console_error_panic_hook` crate provides better debugging of panics by
+# logging them with `console.error`. This is great for development, but requires
+# all the `std::fmt` and `std::panicking` infrastructure, so isn't great for
+# code size when deploying.
+console_error_panic_hook = { version = "0.1.6", optional = true }
+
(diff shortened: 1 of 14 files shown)


Keats:
That's amazing thanks!
I'll handle the CSS and CI don't worry about it

docs/tera-playground/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+# allocator, however.
+#
+# Unfortunately, `wee_alloc` requires nightly Rust when targeting wasm for now.
+wee_alloc = { version = "0.4.5", optional = true }
+tera = { path = "../../", default-features = false }
+serde_json = "1.0.63"
+
+[dev-dependencies]
+wasm-bindgen-test = "0.3.13"
+wasm-bindgen = { version = "0.2.63", features = ["serde-serialize"] }
```
pickfire, line 33: Side note, I don't think this is needed anymore, I just realized this.

PR #609, opened 2021-02-28
Work on playground

Keats:


Keats:
cc @pickfire
Slightly better design but still not great - can't be arsed to spend too much time on design right now.

pickfire:
Why do you have to add the `pkg/` files? Aren't those generated?

docs/webpack.config.js
```diff
@@ -1,10 +1,10 @@
...
 
 module.exports = {
+  mode: "production",
   entry: "./bootstrap.js",
   output: {
-    path: path.resolve(__dirname, "static"),
+    path: path.resolve(__dirname, "static", "playground"),
     filename: "bootstrap.js",
   },
-  mode: "development",
```
pickfire, line 9: We should keep this as development, for production, you can just use `npm run build -- --production` in CI.

PR #617, opened 2021-03-27
Playground remove wasm-bindgen dev-dependencies

pickfire:
I still can't seemed to run the playground with zola

```
Error importing `index.js`: TypeError: WebAssembly: Response has unsupported MIME type '' expected 'application/wasm'
```

### DIFF
--- a/docs/tera-playground/Cargo.toml
+++ b/docs/tera-playground/Cargo.toml
@@ -30,7 +30,6 @@ serde_json = "1.0.63"
 
 [dev-dependencies]
 wasm-bindgen-test = "0.3.13"
-wasm-bindgen = { version = "0.2.63", features = ["serde-serialize"] }
 
 [profile.release]
 # Tell `rustc` to optimize for small code size.


pickfire:
I think tera error handling may need to be improved.

```
Failed to render '__tera_one_off'
```
When the `username` within json is changed to `name`, this would make it hard to figure out what is the issue, but maybe that could be another issue.

Keats:
> I think tera error handling may need to be improved.

We should display all the errors, I think we might stop at the first one of the chain in the playground

pickfire:
First one of the chain? We have error chain or something like that?

Keats:
Something similar, but using the Error::cause field

pickfire:
```rust
#[wasm_bindgen] 
pub fn render(template: &str, context: String) -> Result<String, JsValue> { 
    tera_render(template, &context).map_err(|e| e.to_string().into())
}
```

So the `to_string()` is wrong? I don't get what you mean by `Error::cause` there.

And I still don't get it by reading https://docs.rs/tera/1.7.0/tera/struct.Error.html, never tried `Debug` and `Display` but shouldn't `ToString` works as well?

Keats:
You would need to do something like https://github.com/getzola/zola/blob/master/src/console.rs#L112-L122 because the current Error struct doesn't have a good fmt::Display implementation (I wanted to be able to colour lines differently depending on error types but it's not a good idea in the end).

pickfire:
Looks quite troublesome, maybe I will ignore the error, not sure if I want to do that since I think a better approach is to change `Display` rather than using the internal error representation to find stuff.

Keats:
> I think a better approach is to change Display rather than using the internal error representation to find stuff.

I agree.

Are you getting `Error importing `index.js`: TypeError: WebAssembly: Response has unsupported MIME type '' expected 'application/wasm'`?
I'm not sure why I'm getting this either

pickfire:
> Are you getting Error importing index.js: TypeError: WebAssembly: Response has unsupported MIME type '' expected 'application/wasm'?

Yes, I think that is because zola does not support wasm type. `python -m http.server` correctly shows the MIME type.

Keats:
ooooh yeah that's correct there was this bug. It has already been fixed in the `next` branch of Zola

PR #620, opened 2021-04-02
Allow iteration on a string

Stepland:
I wanted to add some silly per-character rainbow text for my own blog with this style
image

I managed to do so with `split` but it feels hacky
```jinja2
{%- macro rainbow(text) -%}
{%- set color_index = 0 -%}
{%- set text_length = text | length -%}
{%- set text_array = text | split(pat="") | slice(start=1, end=-1) -%}
{%- for i in range(end=text_length) -%}
    {%- set char = text_array | nth(n=i) -%}
    {%- if char != " " -%}
        <span class="color-{{ color_index % 6 }}">{{ char }}</span>
        {%- set_global color_index = color_index + 1 -%}
    {%- else %} {% endif -%}
{%- endfor -%}
{%- endmacro input -%}
```
Adding the ability to iterate on characters with a filter reduces it to
```jinja2
{%- macro rainbow(text) -%}
{%- set color_index = 0 -%}
{%- for char in text | chars -%}
    {%- if char != " " -%}
        <span class="color-{{ color_index % 6 }}">{{ char }}</span>
        {%- set_global color_index = color_index + 1 -%}
    {%- else %} {% endif -%}
{%- endfor -%}
{%- endmacro input -%}
```

I feel like tera not allowing strings to be iterated on directly was intentional so I chose to implement it as a filter. Using filters also allows for other ways of iterating on a string to be implemented, even though I can't think of another useful way to iterate on a string right now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,8 @@ pest_derive = "2.0.2"
 lazy_static = "1.0"
 # used in striptags & titles filters. Already pulled by globwalk
 regex = "1.0"
+# used in truncate filter and string iteration
+unic-segment = "0.9"
 
 # used in slugify filter
 slug = {version = "0.1.1", optional = true}
@@ -31,8 +33,6 @@ humansize = {version = "1", optional = true}
 # used in date format filter
 chrono = {version = "0.4.1", optional = true}
 chrono-tz = {version = "0.5", optional = true}
-# used in truncate filter
-unic-segment = {version = "0.9", optional = true}
 # used in get_random function
 rand = {version = "0.8", optional = true}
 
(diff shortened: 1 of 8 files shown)


Keats:
> I feel like tera not allowing strings to be iterated on directly was intentional 

Not really, more like an oversight. If it works in Jinja2, it should work here as well in this case.

Keats:
@Stepland are you interested in implementing iteration on strings?

Stepland:
I'll try !

Stepland:
I tried repurposing `ForLoopValues::Array` for this but I couldn't figure out how to make it work without turning `ForLoopValues::current_value()` into a big mess, so instead I made a new enum member

Keats (changes requested):
LGTM but can you add some tests?

Keats:
Actually wait, using `.chars` will not work as expected. You want to make the `unic-segment` non optional and use that (see the `truncate` filter for an example).
See https://doc.rust-lang.org/std/primitive.char.html#representation for details

Stepland:
Nice catch !

I added only one test. I *think* it is enough to check that string iteration gives back grapheme clusters and not something else like codepoints. I feel like adding more tests would just result in testing `unic-segment` by proxy.

Keats (approved):

Keats:
Thanks,  I'll merge/release it next week!

Keats:
I should stop giving time estimates haha
Thanks, I'll release it now

PR #621, opened 2021-04-10
Fix parsing of filter arguments separated by whitespaces

gdzx:
The filter_tag rule is compound atomic, denying implicit whitespaces in
sub-rules such as fn_call. This commit marks fn_call as non-atomic, and
adds filter and macro unit tests for arguments separated by spaces.

### DIFF
--- a/src/parser/tests/lexer.rs
+++ b/src/parser/tests/lexer.rs
@@ -468,9 +468,28 @@ fn lex_block_tag() {
 }
 
 #[test]
+fn lex_filter_tag() {
+    let inputs = vec![
+        "{%- filter tag() %}",
+        "{% filter foo(bar=baz) -%}",
+        "{% filter foo(bar=42) %}",
+        "{% filter foo(bar=baz,qux=quz) %}",
+        "{% filter foo(bar=baz, qux=quz) %}",
+        "{% filter foo ( bar=\"baz\", qux=42 ) %}",
+    ];
+    for i in inputs {
+        assert_lex_rule!(Rule::filter_tag, i);
+    }
+}
+
+#[test]
 fn lex_macro_tag() {
-    let inputs =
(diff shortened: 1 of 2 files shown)


gdzx:
For example:

```
{% filter foo(bar=42, baz="qux") %}
{% endfilter %}
```

Returns:

```
ERROR: Failed to parse '__tera_one_off'
    
    Caused by:
         --> 4:22
          |
        4 | {% filter foo(bar=42, baz="qux") %}␊
          |                      ^---
          |
          = expected an identifier (must start with a-z)
```

Keats:
I'm surprised this issue never popped up before :o

src/parser/tests/lexer.rs
```diff
@@ -467,10 +467,29 @@ fn lex_block_tag() {
...
+
 #[test]
 fn lex_macro_tag() {
-    let inputs =
-        vec!["{%- macro tag() %}", "{% macro my_block(name) -%}", "{% macro my_block(name=42) %}"];
+    let inputs = vec![
+        "{%- macro tag() %}",
+        "{% macro my_block(name) -%}",
+        "{% macro my_block(name=42) %}",
+        "{% macro foo ( bar=\"baz\", qux=42 ) %}",
```
Keats, line 491: so Ideally space would be allowed inside the parentheses but not between the name and the `(`

gdzx: Unless I'm mistaken, this last test case passes even without this commit (`macro_fn_wrapper` is non-atomic, I suppose that also applies to `macro_fn`):

```
macro_fn        = _{ ident ~ "(" ~ macro_def_args? ~ ")" }
macro_fn_wrapper = !{ macro_fn }
```

Keats: ah nevermind then!

PR #622, opened 2021-04-13
Missing jinja2 tag in documentation for for-else

MathiasPius:
Causing missing syntax highlighting in the last example here: https://tera.netlify.app/docs/#for


Keats:
Thanks!

Issue #590, opened 2021-01-01
URLs in HashMap keys results in "Failed to render"


joepio:
First off: thanks for making & maintaining this! Love it :)

It seems like the square bracket syntax does not support (some of the characters in) URLs.

```rust
    let mut tera_context = tera::Context::new();
    let mut map= HashMap::new();
    map.insert("https://example.com/url", "url value, not working");
    map.insert("simplekey", "simple key value, working");
    tera_context.insert("map", &map);
```

```
    {{ map["simplekey"] }} => works fine 
    {{ map["https://example.com/url"]}} => "Failed to render \'map_example.html\'" 
```

It took me a while to find the cause of this issue - the error wasn't clear. (but error handling isn't what this issue is about)

Keats:
I think that's because the map key in the parser doesn't support characters like `:` or `/` maybe? A bit weird and shouldn't be too hard to fix if it's that.

MathiasPius:
I did a few tests, and it's actually the `.` that's the problem.

The map index is interpreted as a JSON Pointer into the context when rendered, but the dot is not escaped (since it is valid JSON Pointer notation), which means that `map["https://example.com"]` actually performs a lookup against `map["https://example"].com`.

It appears that serde (probably in accordance with the json spec) makes no distinction between `{ "google.com": 3 }` and  `{ "google": { "com": 3} }`, but I don't know for sure.

It would appear to be a general issue with any string containing dots.

Here's the test I used:
```rs
#[test]
fn test_render_map_with_complex_key() {
    let mut my_tera = Tera::default();
    my_tera
        .add_raw_template("test", r#"{{ map["a.b.c"] }}"#)
        .unwrap();

    let mut map= HashMap::new();
    map.insert("a.b.c", "success");

    let mut tera_context = Context::new();
    tera_context.insert("map", &map);

    my_tera.render("test", &tera_context).unwrap();
}
```

PR #623, opened 2021-04-14
Allow dots in context object key names

MathiasPius:
See: #590

To summarize the change, I'll compare the indexing flow from before to now:

Before:
`{{ map["a.b"] }}` would be turned into a 'key' `map.a.b` which would then get translated into json pointer notation `map/a/b`and then used to retrieve the value.

Now:
`{{ map["a.b"] }}` is turned into the key `map."a.b"` which then gets translated into json pointer notation, respecting quoted strings as a single path segment `map/a.b`

### DIFF
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -956,8 +956,8 @@ mod tests {
         let mut my_tera = Tera::default();
         my_tera
             .add_raw_templates(vec![
-                ("dots", "{{ map[\"a.b.c\"] }}"),
-                ("urls", "{{ map[\"https://example.com\"] }}"),
+                ("dots", r#"{{ map["a.b.c"] }}"#),
+                ("urls", r#"{{ map["https://example.com"] }}"#),
             ])
             .unwrap();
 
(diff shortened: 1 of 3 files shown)


src/tera.rs
```diff
@@ -951,6 +951,27 @@ mod tests {
         assert_eq!(result, "Hello world");
     }
 
+    #[test]
+    fn test_render_map_with_dotted_keys() {
+        let mut my_tera = Tera::default();
+        my_tera
+            .add_raw_templates(vec![
+                ("dots", "{{ map[\"a.b.c\"] }}"),
```
Keats, line 959: Use raw strings `r#"{{ map["a.b.c"] }}"#` so the example is copy/pastable

Keats (changes requested):
Very nice! Thanks a lot

src/tera.rs
MathiasPius: Haha, it's funny you mention it, I did use raw strings originally but didn't see them used in any of the other examples, so I switched it up.

Switched back to raw strings :)

Keats (approved):

Issue #435, opened 2019-07-23
newline to break


haheute:
Hello, 
I have problems to print content from a MySQL text column with line breaks.
In Adminer the text is shown as this:

```
Animal Alphabets
B is for Bee-Eater
```

I try to render it with this line:
`{{ photo.text | replace( from="\n", to="<br>") }}`

the browser shows this:
`Animal Alphabets B is for Bee-Eater`

when I try this:
`{{ photo.text | replace( from="A", to="<br>") }}`

the browser shows:
```
nimal 
lphabets B is for Bee-Eater
```

It seems as if tera (or diesel?) .. or 'something' .. does not seem to find the `\n`.

Is `\n` the right expression to get the newline / line break character?
How is this usually done with tera?

Keats:
I've added a test with your exact example (https://github.com/Keats/tera/commit/af96fbe3f2f807f603ba9cd896ccb3c532e72282) and it works fine, try adding `| safe` (as in `{{ photo.text | replace( from="\n", to="<br>") | safe}}`.
Which version of Tera are you using? I haven't used anything other than the v1 beta for 6 months so maybe there were some bugs before.

haheute:
I was using version 0.11. 
```
[dependencies]
rocket = "0.4"
rocket_contrib = { version = "0.4", features = ["tera_templates", "diesel_mysql_pool", "helmet"] }
serde = "1.0"
serde_derive = "1.0"
serde_json = "1.0"
chrono = { version = "0.4", features = ["serde"] }
diesel = { version = "1.4", features = ["mysql", "r2d2", "chrono"] }
#tera = "0.11"
tera = "1.0.0-beta.13"
dump = "0.1"
etag = "1.0"
```
Adding `| safe` did not help.
Now trying with 1.0.0-beta.13, which throws errors:
```
error[E0277]: the trait bound `tera::Context: models::_IMPL_SERIALIZE_FOR_Photo::_serde::Serialize` is not satisfied
  --> src/main.rs:41:5
   |
41 |     Template::render("imprint", &Context::new())
   |     ^^^^^^^^^^^^^^^^ the trait `models::_IMPL_SERIALIZE_FOR_Photo::_serde::Serialize` is not implemented for `tera::Context`
   |
   = note: required because of the requirements on the impl of `models::_IMPL_SERIALIZE_FOR_Photo::_serde::Serialize` for `&tera::Context`
   = note: required by `rocket_contrib::templates::Template::render`

```

[...]

PR #626, opened 2021-04-19
Add linebreaksbr builtin

dbrgn:
The name is taken from Django, where the same built-in template filter exists.

Adresses #435.


Keats (approved):

Keats:
Thanks!

PR #628, opened 2021-04-22
Add Context::remove

### LINKED ISSUE #627: Add `Context::remove`
It would be useful to remove a key from an old context.

I'm writing a code to make `Tera` generate a rendered string with an instance of `Context` and store the context to a vector.
This vector value may include a bunch of contexts, so I want to omit unused keys from each context to save data size.

This is an example of my case. `body` of `context` won't be used after rendering, so I want to remove `"body"` from `context`.

```rust
let tera = Tera::new("templates/**/*.html").unwrap();
let mut summaries: Vec<Context> = Vec::new();

for content in contents {
    context.insert("title", content.title);
    context.insert("body", content.body);

    println!("{}", tera.render_str("post.html", &context).unwrap());

    context.remove("body") // <-- I want to exclude huge data.
    summaries.push(context);
}

do_something_with(summaries);
```

Does it make sense? If so, I want to open a PR soon.

Thanks,

yykamei:
Close #627 

This patch allows the caller to remove a key from the context.
The behavior of this new method is borrowed from `BTreeMap`.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -118,6 +118,11 @@ impl Context {
         self.data.get(index)
     }
 
+    /// Remove a key from the context, returning the value at the key if the key was previously inserted into the context.
+    pub fn remove(&mut self, index: &str) -> Option<Value> {
+        self.data.remove(index)
+    }
+
     /// Checks if a value exists at a specific index.
     pub fn contains_key(&self, index: &str) -> bool {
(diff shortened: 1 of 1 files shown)


Keats (approved):

yykamei:
Thanks for your quick review :smiley:

PR #632, opened 2021-05-20
Add Tera::get_template_names()

### LINKED ISSUE #631: Add mechanism to get list of all registered templates
Rocket needs a way to determine when the user has manually added templates to a Rocket-managed instance of `Tera`. To do this, Tera needs to expose a mechanism to get a list of all template names. For example, `handlebars` exports https://docs.rs/handlebars/3.5.5/handlebars/struct.Handlebars.html#method.get_templates.

At present, I am using the hidden `Tera.templates` field, but a semver-compatible solution would be appreciated. Exposing a method `template_names()`, marking the field `pub` without `#[doc(hidden)]`, or simply promising to keep the field `pub` and of the same type would suffice.

SergioBenitez:
Closes #631

### DIFF
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -390,6 +390,28 @@ impl Tera {
         }
     }
 
+    /// Returns an iterator over the names of all registered templates in an
+    /// unspecified order.
+    ///
+    /// # Example
+    ///
+    /// ```rust
+    /// use tera::Tera;
+    ///
+    /// let mut tera = Tera::default();
+    /// tera.add_raw_template("foo", "{{ hello }}");
+    /// tera.add_raw_template("another-one.html", "contents go here");
+    ///
+    /// let names: Vec<_> = tera.get_template_names().collect();
+    /// assert_eq!(names.len(), 2);
+    /// assert!(names.contains(&"foo"));
+    /// assert!(names.contains(&"another-one.html"));
+    /// ```
+    #[inline]
+    pub fn get_template_names(&self) -> impl Iterator<Item = &str> {
+        self.templates.keys().map(|s| s.as_str())
+    }
+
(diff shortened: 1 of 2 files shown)


SergioBenitez:
Test failures are unrelated. It seems `memchr`' requires a newer stable version than 1.36 as pinned in the `pinned` test.

src/tera.rs
```diff
@@ -390,6 +390,28 @@ impl Tera {
...
+    /// tera.add_raw_template("another-one.html", "contents go here");
+    ///
+    /// let names: Vec<_> = tera.get_template_names().collect();
+    /// assert_eq!(names.len(), 2);
+    /// assert!(names.contains(&"foo"));
+    /// assert!(names.contains(&"another-one.html"));
+    /// ```
+    #[inline]
+    pub fn get_template_names(&self) -> impl Iterator<Item = &str> {
+        self.templates.keys().map(|s| s.as_str())
```
Keats, line 412: I think it would be worth `collect()`ing in the function as most people will do it anyway and it's not a function that will be called so many times the allocation would be detrimental

Keats (changes requested):

Keats:
Looks like the new pinned version needs to be 1.41.1

src/tera.rs
SergioBenitez: Why `collect()` in the function when the user can do this themselves? In general, you shouldn't allocate for your users. They may want a `Vec`, or they may want a `HashSet`, or they may simply want the `.count()`.

Keats:
Thanks!

Keats:
It's released in 1.10.0

SergioBenitez:
Thank you!

PR #635, opened 2021-06-02
Improvements to docs

rootkea:


Keats (approved):

rootkea:
Instead of opening separate PRs for documentation fixes I'll be adding all the fixes I can find to this PR and then shall mark this as ready. This way this will not bother you much. :)

rootkea:
I'll open another PR for remaining fixes (if any) to docs.

docs/content/docs/_index.md
```diff
@@ -333,15 +333,16 @@ You can check whether a left side is contained in a right side using the `in` op
...
 
 
 ## Manipulating data
 
 ### Assignments
 You can assign values to variables during the rendering.
 Assignments in for loops and macros are scoped to their context but
-assignments outside of those will be set in the global context.
+assignments outside of those will be set in the global context. Furthermore, assignments
+in for loop are valid until the end of the iteration only.
```
Keats, line 345: of the current iteration

Otherwise it sounds like it's valid during the whole for loop.

docs/content/docs/_index.md
```diff
@@ -304,11 +304,11 @@ The priority of operations is the following, from lowest to highest:
...
 - `and`: true if the left and right operands are true
 - `or`: true if the left or right operands are true
-- `not`: negate a statement
+- `not`: negate an expression
 
-#### String concatenation
+#### Concatenation
 
-You can concatenate several strings/idents using the `~` operator.
+You can concatenate several strings/numbers/idents using the `~` operator.
```
Keats, line 311: function calls as well, provided they return a string or a number

Keats (changes requested):

docs/content/docs/_index.md
rootkea: I was thinking about this. Shouldn't function calls be allowed to replace idents if function returns the result of type same as that of ident?

I mean, function calls should be able to replace variables/idents.

e.g.
````
{{ 5 in [0, 1, 2, 3, 4, 5, 6] }}
````
This should be equivalent to:
```
{{ 5 in range(end=7) }}
```
Does parser v2 handles this? Currently it's a parser error.

If a variable is allowed then function returning that type of result is too. So there will be no need to mention 'function calls too' alongside `idents` in docs.

Keats: > Does parser v2 handles this?

Yep

rootkea: Cool! Then we don't need to mention "function calls as well" since one can always replace idents with function calls so only mentioning "idents" is enough.

Otherwise we'll have to accompany "idents" with "function calls as well" all over the doc.

Keats (approved):

PR #643, opened 2021-06-21
Add spaceless builtin

### LINKED ISSUE #641: spaceless feature from Django
For rustdoc, [we're adopting tera templates](https://github.com/rust-lang/rust/pull/86157#issuecomment-864659727). We try to generate HTML with as little unneeded whitespace as necessary, to keep page sizes small. Currently we're doing that manually; with tera templates we'll be using `{%- -%}`. But we still have to deal with newlines and indenting whitespace that we use purely to make the templates more readable.

Per the linked issue, we'll be adding an empty comment to the end of each line, with the whitespace control characters: `{#- -#}`. That does the job, but it's somewhat tedious and hurts readability.

Evidently [Django has a `spaceless` directive](https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#spaceless) that is HTML tag aware and knows how to strip all whitespace between tags. If we could use that with tera templates it could considerably improve the readability and writeability of our templates.

There's a similar issue for attributes within a tag, and I don't know if other place in the jinja ecosystem have a good solution. Specifically, given a template like:

```
<link rel="stylesheet" type="text/css"
      href="{{static_root_path | safe}}normalize{{page.resource_suffix}}.css">
```

We'd like that to render with exactly one space character between `css"` and `href=`. The end-of-line-comment trick doesn't work because it would concatenate the attributes. We can use a `{{- " " -}}`, but it would be great if there were something similar to `spaceless` for this case.

Urgau:
The name is taken from Django, where the same built-in template filter exists.

Fixes #641.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ serde_json = "1.0.11"
 pest = "2.0.2"
 pest_derive = "2.0.2"
 lazy_static = "1.0"
-# used in striptags & titles filters. Already pulled by globwalk
+# used in striptags, spaceless & titles filters. Already pulled by globwalk
 regex = "1.0"
 # used in truncate filter and string iteration
 unic-segment = "0.9"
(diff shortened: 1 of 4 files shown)


Keats:
Cheers that was much simpler than I thought. I'll wait a few days to see if there are more things to merge and will release a new version.

PR #652, opened 2021-07-13
Don't enable oldtime feature of chrono

kornelski:
Chrono has a default feature `oldtime` that pulls in an outdated crate. `tera` is not using it, so it doesn't need to enable it.


Keats:
I think we need to keep some of the default features looking at the CI failure?

kornelski:
Indeed. I hoped chrono-tz would enable std for you, but you also use the `clock` feature. I've made that explicit.

Keats:
Thanks!

PR #653, opened 2021-07-14
Made docs__sidebar sticky

U-C-S:
Just a small feature for quick Navigation in the docs.......

image


Keats:
Very useful, thanks!

PR #662, opened 2021-08-25
escape_html: stop encoding slash, change apostrophe encoding to &apos;, better docs

chris-morgan:
More details in the commit message.

I propose this as an alternative to #568; I wrote it before checking if someone had already done much the same thing (or else I would probably not have bothered—though I also have a vague feeling that I knew about #568 last year but have forgotten about it since), but I think my patch is more thorough.

I also propose that this shouldn’t be considered a breaking change; anything that depended on slash being escaped as `&#x2F;` was already catastrophically wrong; if you want to compare equality of serialised HTML (a very dubious operation in general, tests are the only place I’d even *consider* countenancing it), you need to normalise it first to be *correct*. I suppose this is the main argument against performing this change in a minor or patch release: you may break some tests, whether you consider those tests Incorrect-with-a-capital-I or not. After writing this out I guess I’m more sympathetic to the breaking-change classification, though I’m still fed up with gratuitously mangled URLs! 🙂

/cc @89z


89z:
I have to say, its kind of annoying that youve opened a SECOND pull request, on what is essentially a one line change. Honestly until now I had forgotten about my PR, and even the fact that Tera/Zora even exist. I did so because the development pace has been frustratingly slow, and because of problems with getting a Rust native SASS.

Coupled with the fact that my original request has been ignored for several months, doesnt make me happy to remember any of this. Please dont mention me any more on anything to do with this repo.

Keats:
It hasn't been ignored though? Just that it will be merged for a future major version as it's breaking imo. 
When a behaviour is specifically indicated in the docs, changing it is a breaking change.

PR #665, opened 2021-09-03
Fixed the TODO merge 2 loops into one in Tera and fixed some uneeded borrows

genusistimelord:
I have combined 2 loops into a single loop however this can only be accomplished currently using a Clone of the templates hash map. I also went ahead and fixed some of the Needless borrow Warnings.


Keats:
Can you split the first commit (loop merging) in its own PR?

genusistimelord:
> 
> 
> Can you split the first commit (loop merging) in its own PR?

Yes I can.

genusistimelord:
closing to switch to one that only has Dual loop fix.

Issue #656, opened 2021-07-22
add Context-scoped functions


moschroe:
I use tera in an actix-web project and wanted to expose the [`url_for` method](https://docs.rs/actix-web/3.3.2/actix_web/web/struct.HttpRequest.html#method.url_for) of `HttpRequest` to the template.

The template engine is initialised globally, and so are its registered functions. `url_for` is specific to the request instance. And the latter is neither `Serialisable` nor `Send` nor `Sync`.

I managed to introduce functions local to a `Context`. I had to create a separate trait that does not require `Send+Sync` (in my use case, the `HttpRequest` contains an `Rc`, which is neither). I had to add plumbing to the `Callstack`. The `Processor` was modified so that a Context-local function will shadow a global one.

For all the changes, see: https://github.com/moschroe/tera/commit/c613ad7dc81eddaff0000ecca76d4693b3b9ac2f

An example using these modifications might look like this:
```rust
pub fn register_url_for(context: &mut Context, req: HttpRequest) {
    let closure = move |args: &HashMap<String, Value>| -> tera::Result<Value> {
        return args
            .get("name")
            .and_then(|v| v.as_str())
            .ok_or_else(|| tera::Error::msg("missing 'name'"))
            .and_then(|key| -> tera::Result<Value> {
                let elements = args
                    .get("elements")
                    .and_then(Value::as_array)
                    // TODO: remove Vec allocation by using iterator directly
                    .map(|arr| arr.iter().map(Value::to_string).collect::<Vec<_>>())
                    .unwrap_or(vec![]);
                req.url_for(key, elements)
                    .map_err(|err_url| {
                        tera::Error::msg(format!("unable to create URL: {:?}", err_url))
                    })
                    .map(|url| Value::from(url.as_str()))
            });
    };
    context.register_function("url_for", FunctionRelaxedSafe::from(closure));
}
```

Would this be feasible as a stop-gap solution until 2.0 allows other work-arounds? Shall I open a PR?

Keats:
Wait the `url_for` of actix is on a request....?
How would you change the global Tera types to fit that usecase for v2?

moschroe:
TBH, I have no idea about any possible internals of 2.0?

But seeing the `Sync+Send` on `Function`, for example, that enables `lazy_static!` to work and also obviates pervasive lifetime tracking, I'm not sure how feasible it would be to drop those entirely.

[...]

PR #666, opened 2021-09-05
implement Context-local functions and ergonomic wrappers marking functions as safe

moschroe:
PR as asked for in #656


Keats:
Looks like the pinned CI Rust version needs to be updated

PR #667, opened 2021-09-07
Reduced Dual loops to a single loop in tera::build_inheritance_chains

genusistimelord:
Changed Dual loop into a Single Loop as per a TODO statement within the code.


genusistimelord:
I think the Pin Test needs to be updated.....

PR #671, opened 2021-09-26
Added `get_or_null` filter.

UebelAndre:
In cases where the context used for rendering has fields [skipped when serializing](https://serde.rs/attr-skip-serializing.html), safely determining what to render can be tough as accessing the fields directly or using the existing [get](https://github.com/Keats/tera/blob/f4c855a1452a6b0c5555b83fbc96ed517f957b62/docs/content/docs/_index.md#get) function will result in an error.

This PR introduces `get_or_null` which can be used interchangeably with `get` but will not result in an error. This can allow templates to safely account for these fields.


Keats:
Sounds like this should be an argument to the `get` method instead?

UebelAndre:
It could be but figured that'd be a breaking change. Though I suppose it could be an optional argument. Just hadn't seen one before

Keats:
Here's an example: https://github.com/Keats/tera/blob/master/src/builtins/functions.rs#L100-L111

UebelAndre:
What would the optional field be called?

Keats:
I'm not sure, `allow_null`?

UebelAndre:
After playing around with some Jinja, it seems the behavior here is non fatal there. Was it an intentional decision to have `get` fail instead of returning `null`? I think it might be worth updating the behavior to match jinja.

Keats:
Tera is a bit more strict than Jinja2 on a few things, that's one of these things. I don't think it makes sense to change the behaviour

UebelAndre:
> Tera is a bit more strict than Jinja2 on a few things, that's one of these things. I don't think it makes sense to change the behaviour

I can understand having restrictions for performance reasons, but I feel like this wouldn't have a big impact, if any. While I think it's fine that there's a difference in behavior, this was something that felt kinda needlessly restrictive. Can you expand a bit on why you don't think it makes sense?

UebelAndre:
@Keats I've updated the PR to what I hope is the cleaner approach. I found I was using `get_or_null(...) | default(...)` a bunch so why not combine them? Users could optionally use `get(key="", default=null)` to achieve `get_or_null`. There's also no change in behavior for existing templates.

Keats:
> While I think it's fine that there's a difference in behavior, this was something that felt kinda needlessly restrictive. Can you expand a bit on why you don't think it makes sense?

`get(...)` is meant to be used for keys that don't work in templates so `hey.inexistant` and `hey | get(key="inexistant)` should have the same behaviour. Jinja2 will silently make that `none` and I like that Tera errors instead.

UebelAndre:
@Keats The build failure seems unrelated to the changes I made here.... Can you confirm?

Keats:
Failure is unrelated yes, the minimum rust version just needs to be bumped

docs/content/docs/_index.md
```diff
@@ -1229,6 +1229,9 @@ Marks a variable as safe: HTML will not be escaped anymore.
 Accesses a value from an object when the key is not a Tera identifier.
 Example: `{{ sections | get(key="posts/content") }}`
 
+The `get` filter also has a `default` parameter which can be used to provide a return value when the `key` parameter is missing from the set being filtered.
+Example: `{{ sections | get(key="posts/content", default=null) }}`
```
Keats, line 1233: does that work? I don't think Tera itself has the concept of null in the template? It's something that should be fixed in v2 though

src/builtins/filters/object.rs
```diff
@@ -50,4 +59,32 @@ mod tests {
...
+
+    #[test]
+    fn test_get_or_null_filter_doesnt_exist() {
+        let mut obj = HashMap::new();
+        obj.insert("1".to_string(), "first".to_string());
+        obj.insert("2".to_string(), "second".to_string());
+
+        let mut args = HashMap::new();
+        args.insert("key".to_string(), to_value("3").unwrap());
+        args.insert("default".to_string(), Value::Null);
```
Keats, line 85: don't use serde_json `Value` in the tests, that's not what's going to happen in templates, especially since we don't have null

Keats (changes requested):

docs/content/docs/_index.md
UebelAndre: You are correct. I'd defined that in my own code and just forgot it wasn't built in.

UebelAndre:
> Failure is unrelated yes, the minimum rust version just needs to be bumped

I take it that should be done in a separate PR?

Issue #668, opened 2021-09-09
any way to pass context as parameters?


TheTechRobo:
I'm using Rocket and one issue I've noticed with Tera is that I need to make a BTreeMap or HashMap or similar to pass variables. This is a problem because I have mismatched types in that hashmap.

I've tried using Box as per https://www.simonewebdesign.it/rust-hashmap-insert-values-multiple-types/, but that doesn't seem to be serialisable. I also don't really understand emums. Can I pass the variables as parameters? Thanks!

Keats:
The context takes anything that impl Serialize.

Keats:
Closing as part of the Tera v2 release. If this is still relevant in v2, please reopen — see the 
  migration guide: https://github.com/Keats/tera/blob/master/MIGRATION.md

Issue #605, opened 2021-02-23
Remove 'static lifetime from custom functions


ghraelo:
Currently `Tera.register_function()` requires the function to have a `'static` lifetime. It would be nice if this wasn't the case, so you could use e.g. closures with borrowed data.

Keats:
Closing as part of the Tera v2 release. If this is still relevant in v2, please reopen — see the 
  migration guide: https://github.com/Keats/tera/blob/master/MIGRATION.md

Issue #543, opened 2020-08-01
Access tera context from functions code


technic:
Hi!

I have a situation, where it would be useful to have a global variable in the functions. I faced it when working with translations / localization using fluent. 

Basically on the rust side I have something like this:

```rust
let mut ctx = Context::new();
ctx.insert("LANG", "en");
// more inserts
tera.render(template_name, ctx)
```

And then in the template code I have

```html
<p>translate("some text", lang=LANG)<p>
<p>translate("more text", lang=LANG)<p>
```
e.t.c.

It works, but I have to put `lang` argument every time, which is a bit verbose for my taste.
The only solution that we have for now is to modify `translate` function in the rust code, so it will capture a `lang` variable.
The problem is that [`Function` trait](https://docs.rs/tera/1.4.0/tera/trait.Function.html) is required to be `Send + Sync`.
So, we need to use an external variable and the synchronization for this. Synchronization will mean, that we can not render templates on web server in parallel. Because each client needs different value of the global variable. (for example different languages).

Can't we pass reference to Context inside each function? This mean that lambdas that we use in `register_function` in addition to HashMap with arguments also accepts `&Context `reference. Then I can know value of "LANG" inside translate implementation. This value will be specific to each `render()` call and can work in parallel.

Keats:
It's something that would be nice but is a breaking change. Probably for the next version.

MightyPork:
A similar issue comes up with URLs in `actix-web`. 

There is a `ResourceMap` struct with URL templates its a URL building function `url_for`, but it cannot be serialized.

- One poor man's solution would be to extract the URL templates to something serializavle, inject that into Context, and then use them in a Tera function - like the LANG string above. 
- A better way would be to have some way to pass arbitrary references to the Tera context that can be only accessed from functions and filters. In C this would be a void*, but I'm not really sure how to go about that in rust...
- Or context could allow defining ad-hoc closure functions that can be called by a given name from the template. This may be the best option, if it's possible...

Keats:
I don't know about `ResourceMap` but if it's created outside of Tera, you can just impl the TeraFn trait on it. That's how I pass things in Zola: https://github.com/getzola/zola/blob/master/components/templates/src/global_fns/mod.rs#L24-L46

MightyPork:
@Keats If I read that right, you still have to pass the lang argument to the function, right?

Here's one way to do what I want right now, with pre-generating the URLs and passing them inside the template as strings.

[...]

PR #681, opened 2021-10-18
Add Sync + Send to FunctionRelaxed

CaptainSpof:
34435db introduced `cannot be sent between threads safely` error.

**Disclaimer** I have no idea what I am doing, feel free to disregard this PR.

I've noticed a regression when upgrading to tera 1.13.0. The error was:

```
`(dyn FunctionRelaxed + 'static)` cannot be sent between threads safely
```

After some digging I identified 34435db responsible for the regression. Following the compiler advises, I've managed to "fix" the issue (as in, the code build and the tests are green)

Again, I'm still new to Rust and I haven't looked at what the code I modified do. So handle with care, I guess.

Might be related to  #458

### DIFF
--- a/src/builtins/functions.rs
+++ b/src/builtins/functions.rs
@@ -9,7 +9,7 @@ use serde_json::value::{from_value, to_value, Value};
 use crate::errors::{Error, Result};
 
 /// The context-local function type definition
-pub trait FunctionRelaxed {
+pub trait FunctionRelaxed: Sync + Send  {
     /// The context-local function type definition
     fn call(&self, args: &HashMap<String, Value>) -> Result<Value>;
 
@@ -21,7 +21,7 @@ pub trait FunctionRelaxed {
 
 impl<F> FunctionRelaxed for F
 where
-    F: Fn(&HashMap<String, Value>) -> Result<Value>,
+    F: Fn(&HashMap<String, Value>) -> Result<Value> + Sync + Send,
 {
     fn call(&self, args: &HashMap<String, Value>) -> Result<Value> {
         self(args)


CaptainSpof:
Yes, I didn't use `FunctionRelaxed` by myself. I'm pretty sure the problem was due to the fact that I used rayon in my code. I've managed to reproduced the behavior with a minimal code, roughly doing what I intended:

```
use tera::{Tera, Context};

use rayon::prelude::*;

fn do_stuff(tera: &Tera, context: &Context) -> String {
    tera.render("test", context).unwrap()
}

fn main() {
    debug!("tera_test");
    tera.add_raw_template("test", "{{ name }}");
    context.insert("name", "toto");

    let r: Vec<String> = (0..10).into_par_iter().map(|_| do_stuff(&tera, &context)).collect();
    println!("{:?}", r);
}

```

Discarding rayon with a `into_iter()` does not cause problems.

Keats:
Oh then it looks like the change made Context not sync/send anymore, that's unfortunate.
cc @moschroe any idea on how to solve that?

moschroe:
@CaptainSpof
I implemented a POC for a `Context` with variable guarantees: https://github.com/moschroe/tera/tree/feature_variably-constrained-context, could you check whether this would work for you?
For an example use case that will not work without relaxing the `Send` requirement: https://github.com/moschroe/test-scoped-tera

@Keats Should this get a new issue? New PR?

moschroe:
Of course it has to touch everything `Context` touches. But it does not change the API, I tried to keep that stable. However, I think it is not a minor feature. My original use case would otherwise almost rule out the use of tera altogether, if not requiring hideous workarounds. Also, other people did ask for having non-global functions, or functionality that could often be solved by having this available, e.g. #668, #605 or #543.

All the 'overhead' stays internal (unless a consumer opts in with `Context::new_relaxed()`, TODO: docs) and I don't think it is too much in the first place. Rust is a very precise language, so we need to be precise in describing type bounds :shrug:. Maybe nice, more expressive type aliases could be defined?

At the moment I don't see any less complicated way to express all necessary constraints, keep the functionality of having non-`Send` things in a context (allowing whole classes of use cases) and keep a unified API. And I also could not come up with something less invasive, sorry :grimacing:

PR #688, opened 2021-10-28
Optionally Send+Sync Context

moschroe:
As requested, new PR. I added type aliases which make the generated docs way more clear and it helps distinguish the use cases. For any active users of tera, nothing should change at all. The new functionality of having non-static and non-`Send` & non-`Sync` `Context`-local functions is opt-in by deliberately using `Context::new_threadlocal()`.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -392,4 +392,19 @@ mod tests {
         let mut context = Context::new();
         assert_eq!(context.remove("unknown"), None);
     }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


Keats (changes requested):
Can you add some tests to prove the normal context is Send+Sync so it doesn't happen anymore? Looks good otherwise!

Keats:
Thanks, i'll yank 1.13 and release 1.14

PR #704, opened 2022-01-11
Add feature for just the urlencode builtin

msrd0:
This way, I don't have to pull in a bunch of other dependencies that I don't need, but can still have the urlencode filter which I do need.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,8 +43,8 @@ tempfile = "3"
 
 [features]
 default = ["builtins"]
-builtins = ["builtins_urlencode", "slug", "humansize", "chrono", "chrono-tz", "rand"]
-builtins_urlencode = ["percent-encoding"]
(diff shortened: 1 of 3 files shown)


msrd0:
@Keats any chance you can take a look at this PR?

Cargo.toml
```diff
@@ -43,7 +43,8 @@ tempfile = "3"
 
 [features]
 default = ["builtins"]
-builtins = ["slug", "percent-encoding", "humansize", "chrono", "chrono-tz", "rand"]
+builtins = ["builtins_urlencode", "slug", "humansize", "chrono", "chrono-tz", "rand"]
+builtins_urlencode = ["percent-encoding"]
```
Keats, line 47: can you just name the feature `urlencode`?

msrd0: yes, I renamed the feature to just `urlencode`

Keats:
cool, I'll merge it with https://github.com/Keats/tera/pull/711 when both are ready

PR #720, opened 2022-05-16
add Tera::set_glob

ssendev:
Allow to set globs after creation of Tera so that the server can start even when there are template errors because I like the error formatting in my browser better than in the terminal.

### DIFF
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -381,6 +381,16 @@ impl Tera {
         tera.render_str(input, context)
     }
 
+    /// Returns the glob.
+    pub fn get_glob(&self) -> Option<String> {
+        self.glob.clone()
+    }
+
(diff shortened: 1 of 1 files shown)


Keats:
I don't understand the use-case, why would your glob change at runtime?

ssendev:
It doesn't but I need a Tera instance even if there are currently errors so the server can start and when the templates change a fs watcher can call full_reload on that instance. I guess it could also be accomplished with a new constructor that doesn't try to load the templates. But with this i can `Tera::default()` then `set_glob`

ssendev:
I found a way to replace the whole `Tera` instance with a new one so I don't need this anymore so feel free to just close this PR if you think it's unnecessary. I still think it's a gap in the API since `full_reload` seems handicapped when it's not possible to create an instance with errors. 

Thanks for your quick response and best wishes.

Issue #718, opened 2022-04-11
Multiple brackets do not work sometimes


v1olen:
Hi, 

In recent days while developing our internal project with the use of Tera, we've found (probably) a bug connected with expressions including multiple nested parentheses. 

For example, the following template:

```html
<p>{{ ( ( _somePrice.cash / _somePrice.period ) + ( _somePrice.cash / _somePrice.period ) )  }}</p>
```

parsed using the following rust code:

```rust
use tera::{Context, Tera};
use serde_json::{json, Value};

fn main() {
    let json = json!({
        "_somePrice": {
            "cash": 10,
            "period": 3
        },
    });

    let context = Context::from_value(json).unwrap();
    let result = Tera::one_off(include_str!("../trivial.txt"), &context, true);
    println!("{:#?}", result);
}
```

returns an error:

```rust
Err(
    Error {
        kind: Msg(
            "Failed to parse '__tera_one_off'",
        ),
        source: Some(
            Error {
                kind: Msg(
                    " --> 1:31\n  |\n1 | <p>{{ ( ( _somePrice.cash ) + ( _somePrice.cash ) )  }}</p>\n  |                               ^---\n  |\n  = expected an integer, a float, `true` or `false`, an identifier (must start with a-z), a dotted identifier (identifiers separated by `.`), or a square bracketed identifier (identifiers separated by `.` or `[]`s)",
                ),
                source: None,
            },
        ),
    },
)
```

_(There is no line wrapping on GitHub code snippets, so keep in mind you've to scroll horizontally in order to read the entire error message)_

With the following template, everything passes fine:

```html
<p>{{ ( _somePrice.cash / _somePrice.period ) + ( _somePrice.cash / _somePrice.period ) }}</p>
```

Steps to reproduce:

1. Clone `git@github.com:dsplce-co/tera-bracket-poc.git`
2. Run `cargo run --example trivial` or `cargo run --example complex`

Can somebody help with that? Above we've provided a minimal example but we had more real-life templates we wanted to use and it cannot be done 'cause of this issue.

Keats:
The current version parser is pretty limited when it comes to parenthesis. Tera v2 will have a normal full support but no timeline whatsoever.

Keats:
Closing as part of the Tera v2 release. If this is still relevant in v2, please reopen — see the 
  migration guide: https://github.com/Keats/tera/blob/master/MIGRATION.md

PR #730, opened 2022-06-13
better parenthesis support (#718)

anzome:
- overall math expression parehthesis
- better support for the order of math operations signified by the parenthesis


Keats:
I feel like there was a reason it wasn't that way but I can't remember it right now...

anzome:
Any hint can help. I'll check the history around this place in dictionary.

anzome:
> I feel like there was a reason it wasn't that way but I can't remember it right now...

probably that was an issue for the implementation when the comparison expression were not yet moved into separate term

Keats (approved):

PR #741, opened 2022-08-04
Upgrade chrono

arjentz:
Upgrade `chrono` the newly released version. I only ran `cargo test` and `cargo fuzz ...` (to be completely honest; it is still running) in order to check that everything still works.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
         include:
         - build: pinned
           os: ubuntu-18.04
-          rust: 1.46.0
+          rust: 1.56.0
         - build: stable
           os: ubuntu-18.04
           rust: stable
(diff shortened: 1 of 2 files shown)


arjentz:
I am aware of your plans to put every dependency behind its own feature flag. I still think that's a good idea (I am only using a handful), but I just wanted to get this updated as soon as possible.

Keats:
Looking at the updated versions, we should be getting the new version automatically? No need for a change afaik.

arjentz:
You're right, I missed that. However, wouldn't it be good to force the use of a non-vulnerable version?

Keats (approved):

arjentz:
@Keats Thanks for the approval. Apparently the pipeline doesn’t succeed due to a too low MSRV. Do you want me to do anything about that?

Keats:
You can either update it or I will do it

Keats:
That's a hefty bump! Might bump the minor version of Tera when I release that

Issue #612, opened 2021-03-13
Presenting dates only possible in English (e.g. month using %B becomes "March")


pedzed:
Only English seems to be supported for presenting dates.

An example:
```toml
# config.toml
default_language = "nl"
```

```tera
<!-- index.html -->
{{ "2020-03-20" | date(format="%e %B %Y") }}
```
Expected: "20 maart 2021"
Actual: "20 March 2021"

Keats:
The underlying time library (chrono) only supports English locale currently :/

pedzed:
> The underlying time library (chrono) only supports English locale currently :/

In that case, there are three options:

1. Open an issue/create a PR for Chrono to add the feature
2. Switch to another library
3. Hack a solution together as a patch

Ideally, option one or two would be done. But I don't think that can/will be done in the short term, so even option three would be good for me.

Al I need is translating 12 month names. I'd be happy if there's even a hacky solution.

Keats:
Looks like https://github.com/chronotope/chrono/pull/453 is merged? I haven't checked how to use it though.

A hacky version would be to have some JavaScript reformatting the dates automatically.

pedzed:
> Looks like [chronotope/chrono#453](https://github.com/chronotope/chrono/pull/453) is merged? I haven't checked how to use it and whether it's actually complete/working though.

I think "locale" refers to timezone, not language (or timezone and language).

> A hacky version would be to have some JavaScript reformatting the dates automatically.

I was actually hoping for something compile-time.

Keats:
> I think "locale" refers to timezone, not language (or timezone and language).

I should have read that issue first :(

I don't think there is a solution in Rust currently.

kepae:
Actually, If I read it correctly, https://github.com/chronotope/chrono/pull/453 _does_ provide month/day translation. See the outlying documentation on the `unstable-locales` feature. (The feature imports [pure-rust-locales](https://github.com/cecton/pure-rust-locales), which describes itself as "imported directly from the GNU C Library" and contains the translations directly.)

So, there's some hope for under-the-hood localization (including translation) if the project is comfortable with pinning a version of chrono and using `unstable-locales`.

Keats:
I believe that would only happen for a new version, not for v1

Keats:
Closing as part of the Tera v2 release. If this is still relevant in v2, please reopen — see the 
  migration guide: https://github.com/Keats/tera/blob/master/MIGRATION.md

PR #742, opened 2022-08-06
feat(date): locale (#612)

ZettaScript:
Adds optional `locale` argument in `date` filter, including doc and tests.

Notes:
* For some(?) reason, `chrono::NaiveDateTime::format_localized` does not exist, so timestamps without timezone cannot be localized.
* In order to avoid code duplication, `format_localized` is used event when a locale is not provided (using `chrono`'s default `POSIX`). Looking at `chrono`'s code, this should not affect performance.
* Adds `chrono/unstable-locales`, which may be heavy and restrict `chrono` versions. Should this be in a feature `date-locale`?


Keats:
> Should this be in a feature date-locale?

I think so

ZettaScript:
I added the non-default `date-locale` feature.

PR #744, opened 2022-08-08
eval_in_condition: Use safe_eval rather than eval (#738)

### LINKED ISSUE #738: The template's `in` doesn't work when a slash (/) exists due to escaping
Given the Rust + template

```
context.insert("friends", &vec!["https://Bob", "Kate"]);
...
Output: {% if "https://Bob" in friends %}TRUE{% endif %}
```

the output will be blank.

If we instead run:

- `{% if "https://Bob" in ["https://Bob", "Kate"] %}`
- `{% if "https://Bob"|safe in friends %}TRUE{% endif %}`

then the statement works as expected.

---

After [adding a test to ensure I wasn't mad](https://github.com/Smerity/tera/commit/2ccb8785f18cb55d82f0fdc7bf5ab66ec00583b0) I traced the code for running an `in` to `eval_in_condition` which calls `eval_expression` on the `lhs` and `rhs` inputs. As our example has an `ExprVal::String` on the left we end up escaping `https://Bob`. Thus, to run the above correctly we need to instead run:

```
context.insert("friends", &vec!["https://Bob", "Kate"]);
...
Output: {% if "https://Bob"|safe in friends %}TRUE{% endif %}
```

This works as we prevent the `ExprVal::String` from being escaped.

**If this is expected behaviour:** I'll add a section to the [containing docs](https://tera.netlify.app/docs/#containing)

**If it isn't expected behaviour:** I can investigate this issue further (though would appreciate someone with more Tera knowledge as I'm uncertain how problematic making the `lhs` / `rhs` "unsafe" (i.e. skipping the safety) might be)

Smerity:
Fixes #738

- Added two simple tests to ensure there isn't a regression
- The fix is simple though would appreciate a sanity check, specifically that calling `safe_eval_expression` in `eval_in_condition` can't result in safety issues
  - I don't believe it does but a double check from someone more familiarly with the codebase would be appreciated =]

### DIFF
--- a/src/renderer/processor.rs
+++ b/src/renderer/processor.rs
@@ -303,8 +303,8 @@ impl<'a> Processor<'a> {
     }
 
     fn eval_in_condition(&mut self, in_cond: &'a In) -> Result<bool> {
-        let lhs = self.eval_expression(&in_cond.lhs)?;
-        let rhs = self.eval_expression(&in_cond.rhs)?;
+        let lhs = self.safe_eval_expression(&in_cond.lhs)?;
+        let rhs = self.safe_eval_expression(&in_cond.rhs)?;
 
         let present = match *rhs {
             Value::Array(ref v) => v.contains(&lhs),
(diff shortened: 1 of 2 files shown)


Keats:
Thanks!

PR #746, opened 2022-08-08
Add cargo-deny

pinkforest:
- checks licenses, advisories (PR and cron periodically), advisories will not fail PR but cron will annoy about them

Uses Embark Studios cargo-deny GH Action

https://github.com/EmbarkStudios/cargo-deny-action

deny.toml can be adjusted for variety of things

### DIFF
--- a/.github/workflows/cargo-deny-cron.yml
+++ b/.github/workflows/cargo-deny-cron.yml
@@ -1,0 +1,23 @@
+name: cargo deny
+on:
+  schedule:
+    - cron: '0 0 * * *'
+  push:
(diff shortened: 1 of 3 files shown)


Keats:
Do you have a link to something good that the CI found? Just wondering whether it's worth running in CI or not.

pinkforest:
nah there isn't anything atm - I addressed the few license issues - but it's good to have around when/if something popups e.g. license check or injected dependency from some shady source that I would think is the most profitable check

so it will be silent and doesn't take much air otherwise either

PR #756, opened 2022-09-17
Fix wrong type in `get_random` error message.

rxxrc:
Those errors are also reported for numbers overflowing an `i32` (e.g. `get_random(end=9999999999)`) which is also confusing, but less so than saying a boolean was expected.

All the other functions in this file seem to have the correct types reported.


Keats:
Woops, bad copy/paste. We should also use usize instead of i32

PR #761, opened 2022-10-02
Fix example of loop on maps

ptitfred:
I suppose the example was duplicated from the simple for loop and the content block wasn't updated accordingly by mistake :)


Keats (approved):
Thanks!

PR #762, opened 2022-10-04
Fix clippy lints

tqwewe:
Fix all clippy lints.

There's still some warnings about using deprecated items from the pest crate, but I ignored that for now.


Keats (approved):

Keats:
Already fixed in another merged PR, sorry for being slow.

PR #793, opened 2023-01-18
Add Tera::empty() constructor

poelzi:
Add Tera::empty() to create a Tera instance without any loaded templates but filters and test functions loaded.

This is useful if you only use user provided templates but want to use filters.


poelzi:
year. should have had a second look :)

PR #794, opened 2023-01-20
Remove unused [badges] section in Cargo.toml

eguiraud:
As per https://doc.rust-lang.org/cargo/reference/manifest.html#the-badges-section, that section is not used by crates.io anymore.


Keats:
Thanks!

Issue #574, opened 2020-11-14
globwalk does not work with several relative paths


FibreFoX:
While migrating from a NodeJS-based project to a Rust-based project I found Tera as a template-engine, it was ment to be my introduction project to learn Rust.

As I separate my program and the used templates, I wanted to pass a relative path of some upper folder, but Tera did not find them. After some investigation (and local experiments to change `globwalk` with `glob`, which did not work due to unsupported syntax having curly-braces `"examples/basic/templates/**/*.{html, xml}"`), I found out that it only is an issue for path that point to the current or upper-level folders.

Globwalk does not work for paths starting with`./` or `../`, there is already an issue open at the other project: https://github.com/Gilnaa/globwalk/issues/28

Keats:
So it should be fixed upstream then

FibreFoX:
At least there, but until then I think it's worth to get mentioned somewhere, that's why I opened this issue here. Others might wonder too.

eguiraud:
Hi, due to https://github.com/Gilnaa/globwalk/issues/28 mentioned above passing a path starting with `./` or `../` to `Tera::new` is completely broken. It does not look like https://github.com/Gilnaa/globwalk/issues/28 is going to be fixed any time soon (it's been 2 years already and there is no consensus on the direction to take).

Could Tera maybe side-step the issue by switching to `glob` or `globset`?

In my project I'm working around this by taking care to normalize the path I pass to `Tera::new`, but if there is a clear direction for a fix I could maybe contribute it.

Keats:
> Could Tera maybe side-step the issue by switching to glob or globset?

If we can make it work without breaking any existing code, sure.

eguiraud:
> without breaking any existing code

I am not sure how to guarantee that (I am not sure globwalk and globset behave identically for all inputs).

Something simpler to check would be that the change doesn't break any existing test :sweat_smile:

caemor:
Alternatively we could check if the input starts with `./` or `../` and return an error instead of an empty list and document this for the `Tera::new` command.

Keats:
That would still be a breaking change though.
If someone can do a PR we can test things.

eguiraud:
> we could check if the input starts with ./ or ../ and return an error

> That would still be a breaking change though.

Only for users that expect that adding a `./` at the beginning of the path always results in an empty list. I'm not sure why you would want to keep those weirdos happy... :smile:

> If someone can do a PR we can test things.

[...]

PR #797, opened 2023-01-24
Error out if glob passed to Tera::new starts with ./

eguiraud:
This is because globwalk always returns an empty list in that case, which is likely not what users expect.

The corresponding issue in globwalk is
https://github.com/Gilnaa/globwalk/issues/28.

See https://github.com/Keats/tera/issues/574 for more discussion.


eguiraud:
Closing in favor of https://github.com/Keats/tera/pull/799 .

PR #798, opened 2023-01-25
add indent filter

untereiner:


Keats (approved):

tomasklaen:
I'd like to suggest for the `width` argument to be renamed to `indentation`, which accept both string and a number input, where if it's a string, it'll be used as indentation as is, and if it's a number, it'll be converted into that number of spaces (just as it works now), similar to how `JSON.stringify()` indentation parameter works in JS.

That way we could not only use spaces as indentation, but tabs and anything else as well, such as `> `, `* `, `- `, etc...

Keats:
I would rather have just the string and people have to write `"   "` if they want 3 spaces rather than a number

untereiner:
@Keats Here is the requested change. Sorry It was a bit long

Keats:
Happy with it @tomasklaen ?

tomasklaen:
Yup.

PR #799, opened 2023-01-26
Always canonicalize glob paths passed to Tera

eguiraud:
This is to work around an issue in globwalk (https://github.com/Gilnaa/globwalk/issues/28) due to which paths starting with `./` or `../` never match anything. A regression test has been added as well.

This fixes https://github.com/Keats/tera/issues/574.


eguiraud:
The CI complains that `let...else` statements are unstable citing https://github.com/rust-lang/rust/issues/87335, but as far as I can tell they [have been stabilized](https://github.com/rust-lang/rust/pull/93628) already.

eguiraud:
Ah it's because the CI still uses rustc version 1.56.0. I can change the `let...else` to a match statement if using such an old version of `rustc` (from [October '21, apparently](https://github.com/rust-lang/rust/releases/tag/1.56.0)) is intentional.

Keats:
Yep, no need to bump the MSRV for such a small change.

I remember some issues with canonicalizing paths on Windows but my memory is fuzzy :/

eguiraud:
Changed the `let...else` to a `match` :+1:

Keats (approved):

eguiraud:
Hi, is this ready to merge? :eyes:

Keats:
Yes, i just needed to block some time to merge all the PR and do a release

PR #800, opened 2023-02-02
Fix example for defining custom functions capturing external variables.

rangerjo:
It took me quite a while to get this example right. LSP hints were quite confusing and pointed me towards `dyn Box`, however `impl` seems to do the trick.


Keats:
Thanks!

PR #801, opened 2023-02-13
Fix clippy 0.1.68

godofdream:
I fixed plenty of clippy warnings and errors.
I updated the dependencies.
Also I replaced some deprecated functions of the parser, humansize, and chrono.
The only real change is the sorting function, which should be faster now.

### DIFF
--- a/src/builtins/filters/array.rs
+++ b/src/builtins/filters/array.rs
@@ -80,13 +80,13 @@ pub fn sort(value: &Value, args: &HashMap<String, Value>) -> Result<Value> {
     };
 
     let first = arr[0].pointer(&ptr).ok_or_else(|| {
-        Error::msg(format!("attribute '{}' does not reference a field", attribute))
+        Error::msg(format!("attribute '{attribute}' does not reference a field"))
     })?;
 
     let mut strategy = get_sort_strategy_for_type(first)?;
     for v in &arr {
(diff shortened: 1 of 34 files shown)


src/builtins/filters/array.rs
```diff
@@ -79,20 +78,34 @@ pub fn sort(value: &Value, args: &HashMap<String, Value>) -> Result<Value> {
...
-    })?;
+    // make sure the types are identical, and the field exists
+    let first = arr[0]
+        .pointer(&ptr)
+        .ok_or_else(|| Error::msg(format!("attribute '{attribute}' does not reference a field")))?;
+    let discriminant = std::mem::discriminant(first);
+    let misfits = arr[1..].iter().find_map(|item| 
+        {   
+            let a = item.pointer(&ptr).ok_or(Error::msg(format!("attribute '{attribute}' does not reference a field for {item}")));
+            if a.is_err() {
```
Keats, line 89: any reason we don't use `?` above?

Keats:
Let's try to not bump the MSRV though so that means removing the inlined args in format!

godofdream:
what is the current msrv?
I would add it as clippy config

Keats:
https://github.com/Keats/tera/blob/master/.github/workflows/ci.yml#L18

godofdream:
I'm redoing the pullrequest with a cleaned up branch

PR #802, opened 2023-02-14
Fix deprecations & clippy::all

godofdream:
fixes chrono, prec_climber and humansize deprecations
humansize is extended by the binary attribute (for jinja compatibility)
fixes clippy warnings and errors still maintaining MSRV 1.56
uses total_cmp instead of partialOrd implementation of f64

_I can't tell if updating the dependencies forces to use a newer MSRV._
Edit: tested with rust 1.56.0

### DIFF
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -4,7 +4,7 @@ use std::fmt;
 
 /// The kind of an error (non-exhaustive)
 #[derive(Debug)]
-#[non_exhaustive]
+#[allow(clippy::manual_non_exhaustive)] // reason = "we want to stay backwards compatible, therefore we keep the manual implementation of non_exhaustive"
 pub enum ErrorKind {
     /// Generic error
     Msg(String),
@@ -49,6 +49,11 @@ pub enum ErrorKind {
         /// The context that indicates where the error occurs in the rendering process
         context: String,
     },
+    /// This enum may grow additional variants, so this makes sure clients
+    /// don't count on exhaustive matching. (Otherwise, adding a new variant
+    /// could break existing code.)
+    #[doc(hidden)]
+    __Nonexhaustive,
 }
 
 /// The Error type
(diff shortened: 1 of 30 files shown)


Keats:
Looks like it fails

src/errors.rs
```diff
@@ -4,6 +4,7 @@ use std::fmt;
 
 /// The kind of an error (non-exhaustive)
 #[derive(Debug)]
+#[non_exhaustive]
```
Keats, line 7: Technically a breaking change. I would remove that change just in case.

src/parser/tests/lexer.rs
```diff
@@ -116,9 +116,9 @@ fn lex_string_concat() {
     let inputs = vec![
         "'hello' ~ `hey`",
         "'hello' ~ 1",
-        "'hello' ~ 3.14",
+        "'hello' ~ 3.18",
         "1 ~ 'hello'",
-        "3.14 ~ 'hello'",
+        "3.18 ~ 'hello'",
```
Keats, line 121: why those changes?

Keats (commented):
The 3.14 -> 3.18 changes are a bit weird. One thing to change for the `non_exhaustive` + a rebase but that looks ok.

src/parser/tests/lexer.rs
godofdream: Clippy assumes 3.14 is a rounded pi and throws errors. 3.18 is a simple replacement.

src/errors.rs
godofdream: It was non exhaustive before. I don't see why it's a breaking change

Keats: It was faking non-exhaustiveness, you could still match on `__Nonexhaustive`. With the new `#[non_exhaustive]` you can't, and the attribute is not there anymore so it might break some weird `match` people have. Never underestimate the weird code that can happen!

godofdream: I reverted the manual_non_exhaustive change, and added an exception for clippy with reasoning, why we kept our implementation.

godofdream:
I pinned cargo versions for 1.56.0 compatibility and backported total_cmp, as it's a rust 1.62 feature.

godofdream:
@Keats with this MR the Pipelines should work again (pinned cargo versions)

godofdream:
by the way #762 is a duplicate

Keats (approved):
Nice!

PR #803, opened 2023-02-14
implement a faster sort

godofdream:
using unstable_sort without allocating a copy of the object.
it checks if the pointer (json path) is available for each entry, and if the type (string,array,number) is identical.
Additionally the errormessage is more detailed.

I didn't benchmark yet.


src/builtins/filters/array.rs
```diff
@@ -79,20 +78,40 @@ pub fn sort(value: &Value, args: &HashMap<String, Value>) -> Result<Value> {
...
+            if std::mem::discriminant(a) != discriminant {
+                return Err(Error::msg(format!(
+                    "expected same types but got {} and {}",
+                    first, item
+                )));
+            }
+            Ok(())
+        })
+        .collect();
+    misfits?;
```
Keats, line 103: Better to put `?` after the collect.

src/filter_utils.rs
```diff
@@ -2,6 +2,37 @@ use crate::errors::{Error, Result};
...
+                if let Some(b) = b.as_i64() {
+                    return Ok(a.cmp(&b));
+                }
+            }
+            if let Some(a) = a.as_u64() {
+                if let Some(b) = b.as_u64() {
+                    return Ok(a.cmp(&b));
+                }
+            }
+            Err(Error::msg(format!("{a} cannot be sorted")))
```
Keats, line 28: Likely going to error on the MSRV

Keats (changes requested):

PR #804, opened 2023-02-14
replace try_get_value macro with try_get_value generic

godofdream:
Since generics are possible since ~1.51 we can replace the try_get_value macro with a generic.
To not change the speed of the implementation try_get_value is inlined.

I left the macro inplace as deprecated, so we don't introduce a breaking change.

### DIFF
--- a/src/builtins/filters/string.rs
+++ b/src/builtins/filters/string.rs
@@ -11,6 +11,7 @@ use percent_encoding::{percent_encode, AsciiSet, NON_ALPHANUMERIC};
 
 use crate::errors::{Error, Result};
 use crate::utils;
+use crate::utils::try_get_value;
 
 /// https://url.spec.whatwg.org/#fragment-percent-encode-set
 #[cfg(feature = "urlencode")]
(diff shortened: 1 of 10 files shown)


src/builtins/filters/string.rs
```diff
@@ -411,7 +412,7 @@ mod tests {
         assert!(result.is_err());
         assert_eq!(
             result.err().unwrap().to_string(),
-            "Filter `upper` was called on an incorrect value: got `50` but expected a String"
+            "Filter `upper` was called on an incorrect value: got `50` but expected a alloc::string::String"
```
Keats, line 415: sounds like a regression in terms of error messages

Keats (commented):
I'm not 100% sold on that if it shows the fully qualified name

src/builtins/filters/string.rs
godofdream: we could use [tynm](https://github.com/azriel91/tynm)
tynm shortens the Type to e.g. `Option<String>`

would it be acceptable to add tynm as dependency?

Keats:
I think, let's keep it as is for now and we can change how we get variables for v2 (maybe axum like dependency injection?).

PR #805, opened 2023-02-14
add abs filter

godofdream:
adds an abs filter for numbers


Keats:
This `thread_local` thing is going to be annoying...

PR #806, opened 2023-02-14
don't allow float for slicing

godofdream:
don't allow float for slicing, and making sure negative indexes work as intended

### DIFF
--- a/src/builtins/filters/array.rs
+++ b/src/builtins/filters/array.rs
@@ -244,45 +244,49 @@ pub fn map(value: &Value, args: &HashMap<String, Value>) -> Result<Value> {
     Ok(to_value(arr).unwrap())
 }
 
-#[inline]
-fn get_index(i: f64, array: &[Value]) -> usize {
-    if i >= 0.0 {
-        i as usize
-    } else {
-        (array.len() as f64 + i) as usize
-    }
-}
-
 /// Slice the array
 /// Use the `start` argument to define where to start (inclusive, default to `0`)
 /// and `end` argument to define where to stop (exclusive, default to the length of the array)
-/// `start` and `end` are 0-indexed
+/// `start` and `end` are 0-indexed and are allowed to be negative indexed
 pub fn slice(value: &Value, args: &HashMap<String, Value>) -> Result<Value> {
     let arr = try_get_value!("slice", "value", Vec<Value>, value);
     if arr.is_empty() {
         return Ok(arr.into());
     }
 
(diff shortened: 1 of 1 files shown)


src/builtins/filters/array.rs
```diff
@@ -244,45 +244,49 @@ pub fn map(value: &Value, args: &HashMap<String, Value>) -> Result<Value> {
...
-    if end > arr.len() {
-        end = arr.len();
+    if end > len {
+        end = len;
     }
-
+    if end < 0 && len + end >= 0 {
+        end += len;
+    }
+    println!("{} {}", start, end);
```
Keats, line 282: should be removed

src/builtins/filters/array.rs
```diff
@@ -244,45 +244,49 @@ pub fn map(value: &Value, args: &HashMap<String, Value>) -> Result<Value> {
...
     // Not an error, but returns an empty Vec
-    if start >= end {
+    if start >= end || end < 0 {
         return Ok(Vec::<Value>::new().into());
     }
-
-    Ok(arr[start..end].into())
+    #[allow(clippy::cast_possible_truncation)]
+    #[allow(clippy::cast_sign_loss)]
```
Keats, line 288: what does clippy suggest to do?

src/builtins/filters/array.rs
```diff
@@ -244,45 +244,49 @@ pub fn map(value: &Value, args: &HashMap<String, Value>) -> Result<Value> {
...
     if arr.is_empty() {
         return Ok(arr.into());
     }
 
-    let start = match args.get("start") {
-        Some(val) => get_index(try_get_value!("slice", "start", f64, val), &arr),
+    let len = arr.len() as i64;
+
+    let mut start = match args.get("start") {
+        Some(val) => try_get_value!("slice", "start", i64, val),
```
Keats, line 260: sounds like a breaking change if someone was passing a float

Keats (changes requested):

src/builtins/filters/array.rs
godofdream: Float wasn't documented.
I will add a fallback for floats.

godofdream: If we want to keep floats as Index (what seems like a bad Idea, but is understandable for backwards compatibility) we can close this MR

godofdream:
As we want to keep floats as Index (what seems like a bad Idea, but is understandable for backwards compatibility) we close this MR

PR #809, opened 2023-02-14
benchmarking with workflows

godofdream:
using https://github.com/benchmark-action/github-action-benchmark

@Keats you would need to add the GITHUB_TOKEN

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,32 +1,60 @@
 name: Benchmarking
 on:
-  push:
-    branches:
-      - master
+  pull_request:
 
 permissions:
-  contents: write
-  deployments: write
+  contents: read
+  deployments: read
 
 jobs:
   benchmark:
     name: Run Rust benchmark example
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


Keats:
Are GH actions stable enough to be ok benchmarking on them?

godofdream:
Citing: https://github.com/benchmark-action/github-action-benchmark#stability-of-virtual-environment

As far as watching the benchmark results of examples in this repository, the amplitude of the benchmarks is about +- 10~20%. If your benchmarks use some resources such as networks or file I/O, the amplitude might be bigger.

I'm changing the threshold to something like 30% to ensure it won't cause false warnings that often.

However it could make sense to use some free tier VM (e.g. oracle cloud?) and use that as github runner.

Keats:
I'm not expecting many changes happening on Tera v1, it's mostly stable so I don't know if it's worth adding. I'm tempted to use that action for other projects though, I didn't know about it!

godofdream:
Are GH actions stable enough to be ok benchmarking on them? -> I may have found a way around.

We simply benchmark master and MR during the same run, and compare them directly without saving the results.
We only comment on the commit

PR #810, opened 2023-02-15
Faster get_json_pointer

godofdream:
I improved the speed of get_json_pointer by preallocating the String

before:
test bench_get_json_pointer          ... bench:         114 ns/iter (+/- 5)
test bench_get_json_pointer_with_map ... bench:         568 ns/iter (+/- 10)

after:
test bench_get_json_pointer          ... bench:          92 ns/iter (+/- 3)
test bench_get_json_pointer_with_map ... bench:         434 ns/iter (+/- 13)

not a breaking change.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -226,15 +226,17 @@ pub fn get_json_pointer(key: &str) -> String {
     }
 
     if key.find('"').is_some() {
-        let segments: Vec<&str> = iter::once("")
-            .chain(JSON_POINTER_REGEX.find_iter(key).map(|mat| mat.as_str().trim_matches('"')))
-            .collect();
-        segments.join("/")
+        let mut res = String::with_capacity(key.len() + 1);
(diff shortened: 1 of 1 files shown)


src/context.rs
```diff
@@ -224,15 +224,17 @@ pub fn get_json_pointer(key: &str) -> String {
         // to fix https://github.com/Keats/tera/issues/590
         static ref JSON_POINTER_REGEX: regex::Regex = regex::Regex::new(r#""[^"]*"|[^.]+"#).unwrap();
     }
-
+    let mut res = String::with_capacity(key.len() + 1);
```
Keats, line 227: Would doing `key.len() * 2` be better to make sure no reallocation happens?

godofdream: we know the exact size will be max key.len() +1
In case of the map it could be less, not more.

Keats: Yeah nevermind I realised I was wrong after posting the comment

Issue #821, opened 2023-03-31
Macro fails to be defined at top of template file


sdedovic:
I believe this is related to a few other issues but here is the behaviour I have noticed:

## Issue 1 - Using a macro in the same template where it is defined

```jinja
{% macro input(label, type="text") %}
    <label>
        {{ label }}
        <input type="{{type}}" />
    </label>
{% endmacro input %}

<div>
{{ self::input(label="Name", type="text") }}
</div>
```

and try to render it,
**`src/main.rs`**
```rust
fn main() -> Result<()> {
    let tera = Tera::new("templates/**/*.html")?;
	println!("{:#?}", tera.render("test.html", &Context::default())?);
}
```

it will successfully compile the template but error out when rendering it due to the macro definition
**console**
```bash
$ cargo run

Error: Failed to render 'test.html'

Caused by:
    Invalid macro definition: `input`
```

Issue #644, opened 2021-06-22
Macro definition doesn't work in simple template


mbme:
Macro definition doesn't work in simple template, but works if template extends other template:

# This works:
```rust
use tera::{Context, Tera};

fn main() {
    let mut tera = Tera::default();
    let context = Context::new();

    tera.add_raw_template("base", "{% block content %}{% endblock content %}")
        .unwrap();
    tera.add_raw_template(
        "test-macro-extends",
        "
{% extends \"base\" %}

{% macro hello_world() %}
hello world!
{% endmacro hello_world %}

{% block content %}
{{ self::hello_world() }}
{% endblock content %}
",
    )
    .unwrap();

    assert!(tera.render("test-macro-extends", &context).is_ok());
}
```

# This doesn't work:
```rust
use tera::{Context, Tera};

fn main() {
    let mut tera = Tera::default();
    let context = Context::new();

    tera.add_raw_template(
        "test-macro",
        "
{% macro hello_world() %}
hello world!
{% endmacro hello_world %}

{{ self::hello_world() }}
",
    )
    .unwrap();

    assert!(tera.render("test-macro", &context).is_err());
    println!("{:#?}", tera.render("test-macro", &context));
}
```
### Error:
```
Err(
    Error {
        kind: Msg(
            "Failed to render \'test-macro\'",
        ),
        source: Some(
            Error {
                kind: InvalidMacroDefinition(
                    "hello_world",
                ),
                source: None,
            },
        ),
    },
)
```

jmcph4:
I think I'm also encountering related behaviour, although my macro is more complex.

```tera
{% macro comment_tree(comment, depth) %}
    <p>{{ comment.author.username }} at {{ comment.created }} said:</p>
    <p>{{ comment.contents }}</p>
    <hr>

    {% for child in comment.children %}
        {{ self::comment_tree(child=child, depth=depth + 1) }}
    {% endfor %}
{% endmacro comment_tree %}

<p><a href="{{ link.url }}" target="_blank">{{ link.url }}</a> by {{ link.author.username }} at {{ link.created }}</p>
<hr>

{% for comment in link.comments %}
    {{ self::comment_tree(comment=comment, depth=0) }}
{% endfor %}
```

Using context derived from the following JSON blob:

```json
[...]

Issue #637, opened 2021-06-03
Tera v2 wishlist/changes


Keats:
On top of https://github.com/Keats/tera/issues?q=is%3Aopen+is%3Aissue+label%3A%22For+next+major+version%22

### Operator precedence and expressions ✅ 

Parentheses should work everywhere and precedence should make sense. This is mostly already implemented in a new parser (private for now).

Should something like `{{ (get_page(path=“some-page”)).title }}` work? Right now it requires going through `set` which is probably fine imo.

### Better error messages ✅ 

The new parser is hand-written so we can provide detailed error for each type of error. I’m also thinking of spanning all expressions in the AST for better rendering errors but not 100% convinced on that yet, will need to try.

### Whitespace management

Do trim_blocks and lstrip_blocks by default that can be switched on/off as one thing and not two.

### Indexing changes ❔ 

Right now you can access array indices by doing `my_value.0` or `my_value[0]`. This is in line with Jinja2/Django but it feels a bit weird.

How about allowing only `my_value[0]` so array index is the same as dict index and more like all programming languages (except tuple access)? The dot syntax doesn’t work when the index is a variable anyway which is probably the main usage of indexing. It would be a big departure from Django/Jinja2 though.

### call
https://ttl255.com/jinja2-tutorial-part-5-macros/#call-block is a good example of that feature.
I do find the naming in Jinja2 very confusing though so it would probably be called something different

### Investigating valuable ❔ 

https://tokio.rs/blog/2021-05-valuable
This should avoid an awful lot of cloning and would improve performances a lot. Still not clear whether it will work though.

### Improving filters/tests/functions
They should be able to get values from the context implicitely the same way expression do so we don’t have to repeat things like `lang=lang` in Zola for example.
Maybe a first argument which is something like `get_ctx: |key: &str| -> Option<Value>`?
It would be nice if there was a way to define the arguments in a better way than the current macros too.
Related issue: https://github.com/Keats/tera/issues/543

-- 

Any feedback / other items you want to change/

KillTheMule:
> **Any feedback / other items you want to change?**

I'm not yet a tera user, but I gave it a very serious look, because I want runtime templates (using askama for now, which uses compile-time templates). I'm templating latex files, which use brackets (in particular, the curly ones) all over the place, so what I'd need is different delimiters than `{{` etc. Is that something you would consider? It would certainly be feasible to restrict this to 2-char delimiters (askama does this, I'm using `~<` instead of `{{` right now).

[...]

Issue #370, opened 2018-12-17
Unreachable reached


Keats:
From https://github.com/getzola/zola/issues/558

```
thread 'thread 'thread '<unnamed><unnamed><unnamed>' panicked at '' panicked at 'internal error: entered unreachable code: render_node -> unexpected node: Extends(WS { left: false, right: false }, "layout.html")internal error: entered unreachable code: render_node -> unexpected node: Extends(WS { left: false, right: false }, "layout.html")', ' panicked at '/cargo/registry/src/github.com-1ecc6299db9ec823/tera-0.11.20/src/renderer/processor.rs', :internal error: entered unreachable code: render_node -> unexpected node: Extends(WS { left: false, right: false }, "layout.html")/cargo/registry/src/github.com-1ecc6299db9ec823/tera-0.11.20/src/renderer/processor.rs774', :/cargo/registry/src/github.com-1ecc6299db9ec823/tera-0.11.20/src/renderer/processor.rs:77418::
```

Keats:
So it is trying to render an `extends` tag, not sure how we ever get there...

@focusaurus

Do you have any memories on the kind of templates that was causing the issue? The parser should not allow `extends` in a place where it would get rendered and I can't write a template reproducing it.
The easy solution is to remove the `unreachable!` here but that means we let through bugs.

focusaurus:
Oh interesting, yes that makes sense. I am converting a site that uses `pug` templates and pug templates have an `extends` keyword. Let me futz around a bit and see if I can reproduce it now that I know what the code was trying to do.

Keats:
@focusaurus did you manage to reproduce it?

focusaurus:
Presumably it was something related to `{% extends "layout.html" %}` which I do use in my templates, but I can't reproduce it.

sbstp:
I'm getting an issue similar to this
```
{% macro path_item(path) %}
    <span class="path" title="{{ path.path }}">{{ path.filename }}</span>
{% endmacro path_item %}

...

<td>{{ self::path_item(path=match.info.path) }}</td>
```

```
[...]

PR #822, opened 2023-03-31
Allow macro definition in renderable template

sdedovic:
This change refactors the `MacroDefinition` rendering to no-op so that a template may contain a macro definition and still render. The previous behaviour was to error. 

This means I may define a macro in a template and use it with the `self` namespace, as the test shows. It's useful for me to be able to extract some common behaviour in my templates out without having multiple files.

Related issues:
- #821
- #644

### DIFF
--- a/docs/content/docs/_index.md
+++ b/docs/content/docs/_index.md
@@ -663,7 +663,10 @@ A macro is called like this:
 ```
 Do note that macros, like filters, require keyword arguments.
 If you are trying to call a macro defined in the same file or itself, you will need to use the `self` namespace.
-The `self` namespace can only be used in macros.
+The `self` namespace can only be used in macros. Macros must be defined top-level (they cannot be nested in an if,
+for, etc.) and should only reference arguments, not template variables directly.
+
+
 Macros can be called recursively but there is no limit to recursion so make sure your macro ends.
 
 Here's an example of a recursive macro:
--- a/docs/content/docs/_index.md
+++ b/docs/content/docs/_index.md
@@ -636,7 +636,6 @@ You can also provide a list of templates that are checked for existence before i
 ### Macros
 
 Think of macros as functions or components that you can call and return some text.
-Macros currently need to be defined in a separate file and imported to be useable.
 
 They are defined as follows:
 
@@ -650,7 +649,7 @@ They are defined as follows:
 ```
 As shown in the example above, macro arguments can have a default [literal](@/docs/_index.md#literals) value.
 
-In order to use them, you need to import the file containing the macros:
+If a macro is defined in a separate file, you need to import the file containing the macros:
 
 ```jinja2
 {% import "macros.html" as macros %}
--- a/docs/content/docs/_index.md
+++ b/docs/content/docs/_index.md
@@ -662,9 +662,7 @@ A macro is called like this:
 {{ macros::input(label="Name", type="text") }}
 ```
 Do note that macros, like filters, require keyword arguments.
-If you are trying to call a macro defined in the same file or itself, you will need to use the `self` namespace.
-The `self` namespace can only be used in macros. Macros must be defined top-level (they cannot be nested in an if,
-for, etc.) and should only reference arguments, not template variables directly.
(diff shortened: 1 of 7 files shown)


sdedovic:
Just now seeing the discussion here: https://github.com/Keats/tera/issues/370#issuecomment-455483920

My opinion is that forcing separation of the macro can make things less ergonomic. Additionally, it was be trivial to enforce macros **only** at the top level.

In my case, I have a template for rending a recipe and a macro that renders the list of ingredients. I want to call the macro multiple times, because my recipe has a few different sections (e.g. dough, topping, filling). This template uses some shared CSS and only makes sense in my recipe template. I don't want to create a set of files, one for templates, and one for macros, just to achieve this.

sdedovic:
I have updated the grammer, parser, and docs to force macros at the top-level only. This should remove any concern of a macro in a for, if, etc.

sdedovic:
Result of running `cargo test -- --format=terse`
```
running 423 tests
........................................................................................ 88/423
........................................................................................ 176/423
........................................................................................ 264/423
........................................................................................ 352/423
.......................................................................
test result: ok. 423 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.07s

     Running tests/render_fails.rs (target/debug/deps/render_fails-1e0a12e06effa9e9)

running 11 tests
...........
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.02s

   Doc-tests tera

running 4 tests
....
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.06s```

File: docs/content/docs/_index.md [hunk 1]
```diff
@@ -664,7 +663,10 @@ A macro is called like this:
 ```
 Do note that macros, like filters, require keyword arguments.
 If you are trying to call a macro defined in the same file or itself, you will need to use the `self` namespace.
-The `self` namespace can only be used in macros.
+The `self` namespace can only be used in macros. Macros must be defined top-level (they cannot be nested in an if,
```
Keats, line 666: That first sentence is not true anymore

Keats (commented):
That looks ok. I'm not sure what behaviour I want for Tera v2 concerning that.

docs/content/docs/_index.md
sdedovic: Fixed, and reworded a little.

sdedovic:
> That looks ok. I'm not sure what behaviour I want for Tera v2 concerning that.

I'll leave my reply here and also in the #637 thread:

I believe macro definition should only be done:
- the top-level of a file (not in an if/else, for, or block)
- in any file (self, as in this PR), 
- in any location (after usage, i.e. forward reference).

See the next section to understand why.

---

With jinja2 templates you can do the following:
```jinja2
{% if some_val %}
  {% macro foo() %}42{% endmacro %}
{%  endif %}
{{ foo() }}
```

The macro may or may not be defined during rendering based on the truthiness of `some_val`. This is because Jinja takes the template AST and converts it into a Python module (using string concatenation) and the "rendering" is more or less `eval`ing the Python at runtime.

This can't work easily with Rust, for obvious reasons. (Maybe if you turn the Tera template AST into rust code, compile it, and load it dynamically...) anyway, 

This becomes a point of divergence for Tera macros. The current code will load all the macro definitions into one table as a first pass during rendering. Then they are looked up from this table when traversing the AST, and thus we cannot have dynamic, conditional macro definitions. This can be fixed by defining macros during AST traversal with Context in scope, or by extending the multi-pass approach to optimize the AST and generate macro definitions prior to rendering.

In any case, I personally believe this is likely an anti-pattern. Rarely is there a valid use case to define globals dynamically, and even less so when these act like pure functions, or "template snippets" as I see it.

sdedovic:
Hey @Keats I think the CI is failing due to some kind of configuration issues. Haven't looked into it. Is there a way to re-trigger a build or should I look to fix the problem?

Keats:
I think it was just a spurious issue, i've restarted the jobs

Issue #819, opened 2023-03-15
Non existent path causes panic instead of Error result


chupaty:
New versions of tera panic instead of returning Error when passed an invalid glob path.  For example:
[rsgen-avro](https://github.com/lerouxrgd/rsgen-avro/blob/9e3eb0bfd0f7ba7a583bb7f3166fd67c41f4e708/src/templates.rs#L338
)

let mut tera = Tera::new("/dev/null/*")?;

Problem caused by use of unwrap() instead of ? operator, for example:

[tera.rs](https://github.com/Keats/tera/blob/226d0108cdb64c8d056e46c5c9a67a4a4e8549ea/src/tera.rs#L135)
let parent_dir = std::fs::canonicalize(parent_dir).unwrap();

Keats:
That's not good, I'll fix it today unless someone beats me to it

Keats:
I think it makes sense and is consistent for me to return an error if a path doesn't exist? `templates/*` can perfectly be empty if there are no files, however if the folder doesn't exist I'd want to know rather than silently ignoring whatever is given.

vimpostor:
Your opinion is perfectly reasonable, it just doesn't match what most other languages do. In Bash it can be controlled with the `failglob` option, which is off by default. Similarly Python's `glob.glob("/dev/null/*")` returns the empty list instead of failing.

But I don't even disagree with you, maybe it's best to fix it in zola.

T3kla:
I have spent 2 hours searching whi the hell my Zola project wasn't building to GitHub Pages. Turns git wasn't pushing the empty templates folder.

PR #824, opened 2023-04-06
Return empty list on invalid globs

vimpostor:
As discussed in #819, this restores the previous behaviour to not fail on invalid globs, while still keeping the fix from #799.

Also see https://github.com/getzola/zola/issues/2150 and several other downstream repositories, where that change caused regressions.


Keats (approved):

giraffate:
This need to do a git rebase or git merge master to get CI working because of https://github.com/Keats/tera/pull/825.

PR #827, opened 2023-04-19
Fixes some documentation and adds more code examples.

xfbs:
Hey, awesome project!

I've noticed that the rustdoc could use a little polishing! I know you have the website too but I find myself always consulting the rustdocs while I'm working and so I figured I could put a little work into it. 

Does this look good to you? I have not gone through the entire crate, only fixes a few things here and there. Wanted to make sure this makes sense before putting too much time into it.

Cheers,
-Patrick

### DIFF
--- a/src/builtins/testers.rs
+++ b/src/builtins/testers.rs
@@ -49,7 +49,7 @@ pub fn value_defined(tester_name: &str, value: Option<&Value>) -> Result<()> {
     Ok(())
 }
 
-/// Helper function to extract string from an Option<Value> to remove boilerplate
+/// Helper function to extract string from an [`Option<Value>`] to remove boilerplate
 /// with tester error handling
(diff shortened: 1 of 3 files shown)


Keats (approved):
I like it! I wouldn't do more changes for now because v2 might change some APIs even though I expect most of the public API to be very similar.

Keats:
Looks like there is a test to fix?

xfbs:
> I like it! I wouldn't do more changes for now because v2 might change some APIs even though I expect most of the public API to be very similar.

Awesome! I'll hold off on new PRs then, until you guys get to v2? 

> Looks like there is a test to fix?

Looks like it, I think I just fixed it.

Keats:
> I'll hold off on new PRs then, until you guys get to v2?

Yep it would be better. Thanks for that PR!

PR #838, opened 2023-06-30
Upgrade pest dependency to 2.7.0

basbossink-ds:
Due to the fact the pest developers yanked versions `2.6.0` and `2.6.1` `cargo deny` now creates a warning for `tera`:
```
warning[yanked]: detected yanked crate
    ┌─ /var/whack/workspace/source/tandemdrive/edrive/Cargo.lock:139:1
    │
139 │ pest 2.6.0 registry+https://github.com/rust-lang/crates.io-index
    │ ---------------------------------------------------------------- yanked version
    │
    = pest v2.6.0
      ├── pest_derive v2.6.0
      │   └── tera v1.19.0
....
```

This PR upgrades pest to version `2.7.0` fixing the warning.


robjtede:
The solution for you is to `cargo update` since `cargo-deny` only reads lockfiles. Changing the minimum version requirement in lib is not the correct solution, and in fact could be worse for all users if 2.7.0 ever needed yanking, too.

PR #839, opened 2023-07-05
Fix for square bracket identifiers breaking tests

p-ackland:
Feel free to throw out this PR. Seems possible but unlikely that the fix is this simple but I don't know much about what's going on underneath the hood. 

After testing for my uses, this seems to work but I haven't tested any further.


Keats:
It might be that simple. Some syntax was not allowed because the pest version at that time would not compile correctly, it's probably fixed by now. I'd rather focus on v2 (https://github.com/Keats/tera/issues/637#issuecomment-1626354639) than add more things to this version though

Issue #677, opened 2021-10-12
Add option to disable the `get_env` function


WesleyAC:
The `get_env` function could be a security risk, in the case that an app both:

* Allows untrusted users to write templates
* Uses environment variables to store secrets

I don't think that this is a terribly uncommon configuration, and it'd be good to give an option to disable the `get_env` function in cases where one doesn't need it.

In my opinion, this should be disabled by default, but I'll be happy so long as there's some way to disable it.

Keats:
You can just overwrite the `get_env` filter with a no-op one in your Tera instance

alexlren:
Would it make sense to either add `unregister_function` / `unregister_tester` / `unregister_filter` methods? 
It would cover most use cases and it's not even a breaking change. 

For a v2, I wouldn't mind have a config option to just manually call the registers function myself.

Keats:
I won't add it to v1 as I'd prefer not adding too much API that might be removed later.

seancroach:
I just made a pull request to make this functionality. Hopefully, it gets merged, or some built-in solution will soon be available. However, for now, this works as a bit of a hack to "unregister" a filter, function, or tester:

```
struct NoOp {
    name: &'static str,
}

impl tera::Filter for NoOp {
    fn filter(
        &self,
        _value: &serde_json::Value,
        _args: &HashMap<String, serde_json::Value>,
    ) -> tera::Result<serde_json::Value> {
        Err(tera::Error::filter_not_found(self.name))
    }
}

impl tera::Function for NoOp {
    fn call(
        &self,
        _args: &HashMap<String, serde_json::Value>,
    ) -> tera::Result<serde_json::Value> {
        Err(tera::Error::function_not_found(self.name))
    }
}

impl tera::Test for NoOp {
    fn test(&self, _value: Option<&serde_json::Value>, _args: &[serde_json::Value]) -> tera::Result<bool> {
        Err(tera::Error::test_not_found(self.name))
    }
}
```

Keats:
Closing as part of the Tera v2 release. If this is still relevant in v2, please reopen — see the 
  migration guide: https://github.com/Keats/tera/blob/master/MIGRATION.md

PR #845, opened 2023-07-25
feat: add `Tera:new_uninitialized` (#677)

seancroach:
Exposes a new function that creates an instance of Tera without registering the default filters, testers, and functions. If this is merged, the version of `tera` in `Cargo.toml` will have to get its minor version incremented. No tests were added as no new code was created (except for a function declaration); I simply cut and paste the body from the `Default` implementation. I need this functionality and think no-ops as an overwrite are too hacky. I hope others can benefit from this, too.

### DIFF
--- a/src/tera.rs
+++ b/src/tera.rs
@@ -158,6 +158,24 @@ impl Tera {
         Self::create(dir, true)
     }
 
+    /// Create a new instance of Tera with no templates, filters, testers or functions registered.
+    ///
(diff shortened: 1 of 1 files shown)


Keats:
I'm not a fan of the name but I can see the need hmm

seancroach:
That's totally fine. I made this rather quickly, and just made it... literal. Augment it however you'd like; I have no strong opinions on the matter.

PR #857, opened 2023-09-13
Fix escaping of backslashes in `_index.md`

mgeisler:
A single leading and training space character is stripped from [inline code](https://spec.commonmark.org/0.30/#example-330).


Keats:
Thanks!

PR #864, opened 2023-10-19
Update example on https://keats.github.io/tera/

alex-way:
The example on the homepage doesn't match that of the example within the repo's readme. Updating to avoid confusion.


Keats:
I'd rather update the README with the content from the site

PR #866, opened 2023-10-26
Add a "per_render" Context option

harrisonkaiser:
In the case you have a particularly large Context that you can't easily clone, it is useful to have a the ablity to augment the Context on a particular render. By adding an extra context that can be specified on each render call the large Context can be shared with other render calls across multiple threads, thus sidestepping the need to clone and modify the Context for each thread.

I, somewhat aribrary, choose the `per_render_context` to be a fallback context, if the value is not found in the `Context` then look in the `per_render_context`. It might be desirable to check for overlapping keys and error out if they exist.

### DIFF
--- a/src/renderer/call_stack.rs
+++ b/src/renderer/call_stack.rs
@@ -13,24 +13,33 @@ use crate::Context;
 /// Contains the user data and allows no mutation
 #[derive(Debug)]
 pub struct UserContext<'a> {
-    /// Read-only context
+    /// Read-only shared context
     inner: &'a Context,
+    /// Read-only fall back, per render context
+    fallback: &'a Context,
 }
 
 impl<'a> UserContext<'a> {
     /// Create an immutable user context to be used in the call stack
-    pub fn new(context: &'a Context) -> Self {
-        UserContext { inner: context }
+    pub fn new_with_shared_fallback(context: &'a Context, fallback: &'a Context) -> Self {
+        UserContext { inner: context, fallback }
     }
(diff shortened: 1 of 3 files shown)


Keats:
Thanks, I won't add new features to the v1 though. v2 progess is at https://github.com/Keats/tera2 but kinda stalled for a bit due to parental leave.

harrisonkaiser:
> Thanks, I won't add new features to the v1 though. v2 progess is at https://github.com/Keats/tera2 but kinda stalled for a bit due to parental leave.

Thanks for the quick response! Tera is a great tool and v2 looks exciting!

Keats:
Note that in v2, cloning contexts should be cheaper so maybe that specific feature won't be needed?

harrisonkaiser:
> Note that in v2, cloning contexts should be cheaper so maybe that specific feature won't be needed?

Will it be cheap to modify the cloned context? I'd like context to vary slightly on each render.

One option is to use: https://docs.rs/im/latest/im/

Keats:
> Will it be cheap to modify the cloned context?

Not really. Most of the Value will be in Arc so cloning is cheap but modifying it, not so much. You can play around https://github.com/Keats/tera2 to see how it behaves. It's mostly working, except for functions/filters/tests not implemented yet and error reporting not working for rendering yet.

For the context/serialization part, it should be kinda complete.

harrisonkaiser:
Okay, I'll close this PR.

PR #867, opened 2023-10-26
Avoid a Vec allocation

harrisonkaiser:
This was found using clippy with `-D clippy::perf`. Since we know the size of the slice at compile time, we can avoid heap allocating a Vec.


harrisonkaiser:
Interesting: ci / test (msrv) failed:
```
error: package `regex-automata v0.4.3` cannot be built because it requires rustc 1.65 or newer, while the currently active rustc version is 1.63.0
```
regex-automata must have bumped their msrv since the last PR.

Keats:
Thanks!

PR #880, opened 2023-12-11
Rio Templates v0.1.0

jayhale:
Establish Rio Templates by:

- Removing or replacing repository extras (license, changelog, etc.)
- Removing the _Tera_ name
   - `Tera` struct renamed to `Engine` in parallel with other template libraries
- Removing fuzzing (for now)
- Removing benchmarks (for now)
- Removing feature flags (all by default now)
- Update crate metadata


jayhale:
Wrong base by accident - sorry!

PR #886, opened 2023-12-23
Bump globwalk dep to 0.9

Gilnaa:
Bumps globwalk to version 0.9, mostly to get rid of bitflags 1.* transitive dependecy. (see https://github.com/Gilnaa/globwalk/pull/34)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ include = ["/src/**/*", "/LICENSE", "/README.md", "/CHANGELOG.md"]
 rust-version = "1.65"
 
 [dependencies]
-globwalk = "0.8.1"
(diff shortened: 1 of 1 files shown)


Gilnaa:
It looks like [`ignore` 0.4.21](https://crates.io/crates/ignore) has a much higher MSRV due to the new use of `OnceLock`. (available since Rust 1.70.0)
This doesn't look a regression that is related to this PR, (this can be reproduced by checking-out an older version of globwalk and running `cargo +1.65.0 check`).

Not sure what to make of it, honestly.

Keats:
It's the problem of having a MSRV, in practice it can change at any point without a direct dependency update :/
In practice I think I will just wait for v2 to upgrade

Issue #861, opened 2023-10-08
[FEATURE REQUEST] Macros for making Context have less boilerplate


MarcusSanchez:
Make a macro that allows quick Context for easy drop in values. Something like this:
```rust
#[macro_export]
    macro_rules! ctx {
        ({
            $($key:literal : $value:expr),* $(,)?
        }) => {
            {
                let mut ctx = Context::new();
                $(
                    ctx.insert($key, $value);
                )*
                ctx
            }
        };
    }
```
That allows this:
```rust 
    let mut context = Context::new();
    context.insert("Hello", "world");
    context.insert("answer", &42);
    context.insert("person", &person);
 ```
 to turn into:
 ```rust
    let context = ctx!({
        "hello": "world",
        "answer": &42,
        "person": &person
    });
```
The exact syntax for the macro doesn't really matter, anything really will be nice, just not a context.insert() every line, this will also make it easier to pass these as parameters.

Keats:
This will probably be added in v2

GandelXIV:
@MarcusSanchez I wrote this very simple implementation for my personal projects:

```rust
#[macro_export]
macro_rules! context {
    (
        $(
            $key:ident $(=> $value:expr)? $(,)*
        )*
    ) => {
        {
            let mut context = Context::new();
            $(
                context.insert(stringify!($key), $($value)?);
            )*
            context
        }
    };
}
```
Using it this way:
```rust
let ctx = context! {
    name => "Jack",
    age => 24
};
```
Expands to:
```rust
let ctx = {
    let mut context = Context::new();
    context.insert("name", "Jack");
    context.insert("age", 24);
    context
};
```

GandelXIV:
We could also get inspiration from [minijinja's implementation](https://docs.rs/minijinja/latest/minijinja/macro.context.html)

GandelXIV:
Made a pull request #887

Keats:
Closing as part of the Tera v2 release. If this is still relevant in v2, please reopen — see the 
  migration guide: https://github.com/Keats/tera/blob/master/MIGRATION.md

PR #887, opened 2023-12-24
Add context! macro to reduce boilerplate when building Contexts (#861)

GandelXIV:
This macro greatly simplifies building Contexts, as you just pass it key-value pairs and inserts them for you.

Example:
```rust
use tera::context;
let ctx = context! {
    name => "Brian",
    age => &24
};
```

Expands to:
```rust
let ctx = {
    context.insert("name", "Brian");
    context.insert("age", &24);
    context
};
```

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -377,6 +377,23 @@ fn parse_index(s: &str) -> Option<usize> {
 }
 
 /// Creates a context from key value pairs
+///
+/// Example:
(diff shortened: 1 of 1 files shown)


GandelXIV:
Associated issue: #861

GandelXIV:
Sure, here you [go](https://github.com/Keats/tera2/pull/24)

PR #894, opened 2024-01-28
parser: allocate vector with exact capacity

lucab:
This tweaks `parse_content()` logic in order to reduce the amount of vector growing, as the final vector capacity is already known beforehand.

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -1044,9 +1044,10 @@ fn parse_if(pair: Pair<Rule>) -> TeraResult<Node> {
 }
 
 fn parse_content(pair: Pair<Rule>) -> TeraResult<Vec<Node>> {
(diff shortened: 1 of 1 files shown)


lucab:
For reference, `parse_content()` is indirectly recursive (through `parse_block()`) so the compounded effect of this is noticeable in memory patterns of consuming applications.

As an example, on a `zola build` run on the `test_site` this spot accounts for ~2% of all temporary allocations (where `zola_patched` is a local build of the `next` branch):
tera

Keats (approved):

lucab:
MSRV bump to fix CI is at https://github.com/Keats/tera/pull/895.

lucab:
This has been rebased to latest `master` and CI is green, PR is now ready to be merged.

Issue #889, opened 2024-01-12
Round filter does not return int with precision = 0


jasger9000:
When using the `| round` filter with `precision = 0` to return a natural number, the filter still returns a float leading to unexpected behavior when using it in combination with functions expecting an integer.

## Minimal code that does not work
```rust
use tera::{Tera, Context};

fn main() {
    let test = Tera::one_off(
        "{% for i in range(end=5.5 | round) %} {{ i }}\n {% endfor %}",
        &Context::new(),
        true
    );

    print!("{test:?}");
}
```
### Output
Err(Error { kind: Msg("Failed to render '__tera_one_off'"), source: Some(Error { kind: CallFunction("range"), source: Some(Error { kind: Msg("Function `range` received end=6.0 but `end` can only be a number"), source: None }) }) })

Keats:
Closing as part of the Tera v2 release. If this is still relevant in v2, please reopen — see the 
  migration guide: https://github.com/Keats/tera/blob/master/MIGRATION.md

PR #899, opened 2024-02-10
round filter: perform cast to i128 if precision is 0

Raymi306:
For issue #889 

Perform an additional check on precision, and cast to an integer if it is 0.


src/builtins/filters/number.rs
```diff
@@ -60,15 +60,24 @@ pub fn round(value: &Value, args: &HashMap<String, Value>) -> Result<Value> {
...
-            method
-        ))),
+                method
+            )))
+        }
+    };
+
+    // Do we need to return an integer or a float?
+    if precision == 0 {
+        Ok(to_value(intermediate_result as i128).unwrap())
```
Keats, line 78: isn't i128 in serde_json converted to a string?

Raymi306:
I truthfully am not super familiar with serde_json, perhaps i need a better
test for this.  I originally was using i64 but thought that i128 might
better handle some of the big numbers an f64 might represent.

Issue #774, opened 2022-11-06
Add merge filter


mcexit:
Pardon my ignorance as I'm pretty new to Rust and Tera, but I'm not seeing an easy way to merge two hashmaps (similar to Hugo's [merge](https://gohugo.io/functions/merge/)). I envision a deep merge by default or an argument to do a deep merge.

A nice to have would be an argument to merge on a common attribute as well:

`data.toml`
```
[[author]]
name = 'John Doe'
email = 'jon@doellc.com'

[[author]]
name = 'Jane Doe'
email = 'jane@doellc.com'
```

`data variable`
```
[[data.author]]
name = 'Jane Doe'
website = 'https://doellc.com'

[[data.author]]
name = 'John Roe'
email = 'john.roe@roeamusements.com'
website = 'https://www.roeamusements.com'
```

Would result in:

```
{
  "author": [
    {
      "email": "jon@doellc.com",
      "name": "John Doe"
    },
    {
      "email": "jane@doellc.com",
      "name": "Jane Doe",
      "website": "https://doellc.com"
    },
    {
      "email": "john.roe@roeamusements.com",
      "name": "John Roe",
      "website": "https://www.roeamusements.com"
    }
  ]
}
```

Keats:
Closing as part of the Tera v2 release. If this is still relevant in v2, please reopen — see the 
  migration guide: https://github.com/Keats/tera/blob/master/MIGRATION.md

PR #900, opened 2024-02-10
Add a simple merge filter

Raymi306:
For issue #774, although I suspect I am not quite meeting all of the goals set forth here.

With v2 coming up, I'm not sure if this is wanted right now.  I also suspect I could make it fancier and take a list of objects and merge all of them somehow.

Anyway, feel free to discard this, at the very least it is helping me get used to reading and writing Rust again after writing Python all day at work.


Keats:
Yeah let's not add new things to v1. I'm wrondering whether we can make some array/dict operations built in the language rather than going through a file, eg `obj1 ~ obj2` for merging, like we do for string concatenation.

Issue #664, opened 2021-09-02
Extends to be compatible with Include


genusistimelord:
So I am attempting to port some of my old python code that used Jinja2 and i noticed that I can not include something that extends something else. I don't understand why this would be?  But lets say I have a card.html that contains blocks and then several other files that use this as a default template via extends Like  Admin-info-card.html

if i were to load the page info.html which extends a base.html and attempt to include in a card extended component it will not due to includes not accepting extends. 

```
{% extends 'base.htm' %}

{% from 'form_utils.htm' import render_field %}

{% block content %}
	<form method="POST">
		<div uk-grid>
			<div class="uk-width-1-5">
				{% include "components/admin-menu.html" %}
			</div>
			<div class="uk-width-expand">
				{% include "components/admin-info-card.html" %}
			</div>
		</div>
	</form>
{% endblock %}
```

the above is the info.html and below is the admin-info-card.html

```
{% extends "components/card.html" %}

{% block card %}
{% block card_title %}
Admin Information
{% endblock card_title %}
{% block card_body %}
<div uk-grid>
    <div class="uk-width-1-2">
        <span>Newest Member:</span>
        <a href="{{ url_for('admin.user_edit', user_id=newest.id) }}">{{ newest.username }}</a>
    </div>
    <div class="uk-width-1-2"> <span>Total Users:</span> {{ users }} </div>
    <div class="uk-width-1-2"> <span>Total Logs:</span> {{ logs }} </div>
</div>
{% endblock card_body %}
{% endblock card %}
```

Would it be possible to Extend the Extends to work in a way that would replace the current code with the extended bits before including it into the includes or at least have a limitation where you can only have 1 extends per document and the document being included cant contain the same extends as the document including it? Anyways this was a very common use pattern in Flask/jinja

Keats:
Includes in Tera are a bit dumb. I pretty much never use them myself in both Tera and Jinja2 except for some very basic templates so I didn't bother implementing inheritance for them. Let's see if more people want that but I'm inclined to not support that.

genusistimelord:
yeah in a sense i was looking at how you did them and i would say you added them in the quickest possible way. XD 
However Rule of thumb is if you want more people to use Tera over other stuff then more features and support for things jinja and the rest have is a must. I might Attempt to fix the issue myself later on once i read into more on how you did things.

Keats:
[...]

Issue #592, opened 2021-01-12
Inconsistent inheritance when mixing include and blocks


posiczko:
When I create a partial that's included in parent html and that partial contains a block, that block is not being overridden
when in a child overrides it:

1. I create a base.html with:

<!DOCTYPE html>
<html>
  {% include "header.html" %}
</html>

2. I create a header.html with:

<head>
  <title>{% block title %}Title in Header{% endblock title %}</title>
</head>

3. I create article.html with:

{% extends "base.html" %}
{% block title %}
    <title>Article Title</title>
{% endblock title %}

Expected behavior is for title block from article.html to override base.html. It does not. 

This may be related to feature #532 with inheritance breaking macros import. 

Thank you!

CliffHan:
I met this problem too.

Keats:
It's the same yes. The current parser is not very smart and doesn't handle cases like this or macros well. It will be fixed for v2

Keats:
Closing as part of the Tera v2 release. If this is still relevant in v2, please reopen — see the 
  migration guide: https://github.com/Keats/tera/blob/master/MIGRATION.md

Issue #454, opened 2019-09-26
How to extend a template that includes sub-templates?


mpizenberg:
Zola version: 0.8.0

I'd like to have a non-bloated `base.html` template so I made it this way.

```
{# base.html #}
<!DOCTYPE html>
<html>
	{% include "head.html" %}
	{% include "body.html" %}
</html>
```

```
{# body.html #}
<body>
	{% block content %}{% endblock content %}
</body>
```

I though I could override the `content` bloc inside my `about.html` template that extends `base.html` but it appears to not be working, without any compilation issue though.

```
{# about.html #}
{% extends "base.html" %}
{% block content %}
Hello world in about.html
{% endblock content %}
```

My `about.html` page is rendered exactly as my `base.html` page so it appears that we cannot override an "included" block in inherited templates.

It is quite surprising but is that the intended behavior?

Keats:
It is the intended behaviour, blocks are found when parsing the templates, `include` are not embedded in the AST, they are just their own nodes in it.

That's unlikely to be changed soon though

Rylon:
Hi @Keats I just encountered this issue myself. I have the same intention as the OP, to split page sections into 'partials' to keep the templates tidy, for example, my `<head>` section is 36 lines long with the various pieces of metadata.

The Tera docs suggest that the 'current context' is available to everything in the included template, and indeed this works with `set` or `set_global`, but blocks aren't working and logically I would expected them to in this case.

I see your original reply, but I'm wondering if there is another way to achieve this? What we essentially want is a way to tell the engine to treat it as if the contents of the partial were actually in-place where the `include` line is in the parent template.

Thanks for any help you can offer!

Keats:
Closing as part of the Tera v2 release. If this is still relevant in v2, please reopen — see the 
  migration guide: https://github.com/Keats/tera/blob/master/MIGRATION.md

Issue #385, opened 2019-01-26
Error when including a template that extends another


Arnaz87:
Test project showcasing the error: https://github.com/Arnaz87/tera-test

And the traceback:
~~~
thread 'main' panicked at 'internal error: entered unreachable code: render_node -> unexpected node: Extends(WS { left: false, right: false }, "parent-holder.txt")', /home/arnaud/.cargo/git/checkouts/tera-ae1ac1f255426cfc/68dfeb5/src/renderer/processor.rs:812:18
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:59
             at src/libstd/panicking.rs:211
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:227
   4: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:491
   5: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:398
   6: std::panicking::begin_panic_fmt
             at src/libstd/panicking.rs:353
   7: tera::renderer::processor::Processor::render_node
             at /home/arnaud/.cargo/git/checkouts/tera-ae1ac1f255426cfc/68dfeb5/src/renderer/processor.rs:812
   8: tera::renderer::processor::Processor::render_body
             at /home/arnaud/.cargo/git/checkouts/tera-ae1ac1f255426cfc/68dfeb5/src/renderer/processor.rs:150
   9: tera::renderer::processor::Processor::render_node
             at /home/arnaud/.cargo/git/checkouts/tera-ae1ac1f255426cfc/68dfeb5/src/renderer/processor.rs:804
  10: tera::renderer::processor::Processor::render
             at /home/arnaud/.cargo/git/checkouts/tera-ae1ac1f255426cfc/68dfeb5/src/renderer/processor.rs:858
  11: tera::renderer::Renderer::render
             at /home/arnaud/.cargo/git/checkouts/tera-ae1ac1f255426cfc/68dfeb5/src/renderer/mod.rs:54
  12: tera::tera::Tera::render
             at /home/arnaud/.cargo/git/checkouts/tera-ae1ac1f255426cfc/68dfeb5/src/tera.rs:309
[...]

PR #909, opened 2024-05-02
Extends inside include

susonwaiba:
For issues:

- #385
- #370

- #454
- #592
- #664

- And my personal works.

Changes:

- Exposed current context
- if `template.parents` is not empty 
  - Check and throw error, if include template has recursive parent template (to prevent loop)
  - Using Processor
- else
  - render_body (as it was)
- Removed error test
- Removed `Note` from docs

Here, `tera.get_template(tpl_name)` seems to already parse parents if any exists.
Using `Processor` might create separate context ??
Performance ??


Keats:
I don't really want to add that feature though.

PR #919, opened 2024-05-22
Mention whitespace control in expressions and comments

uncenter:
Docs were never added as part of https://github.com/Keats/tera/pull/430.


Keats (approved):
Thanks!

PR #921, opened 2024-05-28
Add pad filter

uncenter:
Adds a new filter called `pad` which pads a given string to a given length with a given fill character (defaults to a space).

### DIFF
--- a/docs/content/docs/_index.md
+++ b/docs/content/docs/_index.md
@@ -786,6 +786,9 @@ Converts a string to uppercase.
 #### wordcount
 Returns the number of words in a string.
 
+#### pad
+Pad a string to a given length with a given fill character.
+
 #### capitalize
 Returns the string with all its characters lowercased apart from the first char which is uppercased.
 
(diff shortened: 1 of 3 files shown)


uncenter:
Ah, just saw https://github.com/Keats/tera/pull/900#issuecomment-1938673386. Is it worth opening a PR over at https://github.com/Keats/tera2 instead or is that premature?

Keats:
Filters are not implemented yet in tera2 but I would wait. I did work a bit on it yesterday and i'm seeing the light regarding that. A dim light still but it's getting there.

uncenter:
Glad to hear it! Excited for tera2. Would you be interested in adding this sort of filter when the time comes for tera2 though? I also have some thoughts about some filters that should be potentially removed/deprecated in the next breaking version...

Keats:
Some will be yes, i'll create a separate crate for non-basic filters/functions so i can do breaking change in that without issues. This one should be able to be added in the crate itself

PR #928, opened 2024-07-10
Backport changes made to tera

cheesycod:
NOTE: This PR is not intended to be merged as-is. Its just a PR of the changes made in case upstream wants to use parts of it for themselves.


Keats:
What are the async things you are doing in the templates?

PR #929, opened 2024-07-15
Adding RenderContext trait so users can provide access to arbitrary structs without converting it all into a Value

brosander:
I think that this will help users expose more extensive data structures to Tera without needing to recursively convert to a `Value` when rendering.

The existing api should still be compatible because both `Context` and any immutable borrow of a `RenderContext` will implement `RenderContext` so the render calls in tera.rs should still work with `&Context`.

I don't think I've changed any logic for the `Context` rendering.  It is all confined to the `RenderContext` trait definition now though.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,7 @@ chrono-tz = {version = "0.9", optional = true}
 rand = {version = "0.8", optional = true}
 
 [dev-dependencies]
+elsa = "1.10.0"
 serde_derive = "1.0"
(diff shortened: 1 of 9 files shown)


Keats:
Nothing like that is going to be merged, all work is happening in https://github.com/Keats/tera2 which is a rewrite from scratch

Keats:
https://github.com/Keats/tera/issues/637#issuecomment-2209530190

PR #941, opened 2024-10-28
feat(docs): submenu open for h3 in docs page

elpiel:
Here's a useful (to me) idea to allow seeing built-ins (filters, fns, etc.) directly in the menu.
One way this could be improved is to remove some headings in the docs  (`For`, `Filters`):

image

here's a view of the Built-ins menu:

image


Keats:
Wouldn't it make the menu too big in practice when opening the filters? Since the docs are a single page I don't think it's a huge pain point to search?

elpiel:
The reason I was checking the documentation in the first place was not so much to see what each filter does, but rather to remind myself of the possible filters.

That's why I said reasonable for me 😅

PR #947, opened 2024-11-20
Allow 'default filter' when (custom) function returns null

casper-embion:
Adds the ability to use the default filter when calling functions.
The default will be applied when the function returns a 'null object'.

example usage:
`
{% set my_var = my_function() | default(value=0) %}
`


Keats:
I think it's ok but it's already working in the next version so I'm not sure I want to add more changes to v1

Issue #925, opened 2024-06-18
Build problem with 1.20.0


fxwiegand:
There seems to be an issue with the build of `1.20.0`:

```
   Compiling tera v1.20.0
error[E0599]: no method named `len` found for struct `Pairs` in the current scope
Error:     --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tera-1.20.0/src/parser/mod.rs:1049:46
     |
1049 |     let mut nodes = Vec::with_capacity(pairs.len());
     |                                              ^^^
     |
help: there is a method `le` with a similar name, but with different arguments
    --> /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/iter/traits/iterator.rs:3887:5

For more information about this error, try `rustc --explain E0599`.
error: could not compile `tera` (lib) due to 1 previous error
```

This happens on our [dependabot PR](https://github.com/datavzrd/datavzrd/actions/runs/9564603290/job/26365726546) that wants to update tera.

fxwiegand:
Solved with updating other crates.

hubcio:
@fxwiegand do you mind releasing a version with fix?

fxwiegand:
Yep that also did the trick for me.

rorynugent:
Any chance you could share what crates were updated in `Cargo.lock` that addressed this issue?
There are scenarios where a blanket `cargo update` for a complex workspace is not an option.

fxwiegand:
> Any chance you could share what crates were updated in `Cargo.lock` that addressed this issue? There are scenarios where a blanket `cargo update` for a complex workspace is not an option.

Sure should have been this commit: https://github.com/alignoth/alignoth/pull/238/commits/c798da1a8cb9d0e51789e78efda3c32ab2f6ea42

rorynugent:
Thanks @fxwiegand! That was very helpful.

I narrowed it down to `pest` and v2.6.0 where the `len()` function was added. In other words, `tera `needs to be run alongside pest of >= v2.6.0.

https://github.com/pest-parser/pest/compare/v2.5.7...v2.6.0#diff-6557813bfc2001db4901f6a0f6b9c6da4631fd4c269a5ad24cc4bb3172a60452R105-R111

PR #951, opened 2024-12-13
pest: update to v2.7

rorynugent:
Addresses an issue when using `tera` as a library where method `len` is not found in Pest.
`tera` specifies pest v2.5.5 but v2.60 or greater is needed. Unfortunately, versions v2.6.0-2.6.1 have been yanked so v2.7 is needed.

Related issue: https://github.com/Keats/tera/issues/925


Keats:
Cargo should already pick up 2.7 automatically without that patch, changing just the minor version doesn't really do anything here

PR #956, opened 2025-01-17
Add default context feature for Context created by Tera

soutade:
Add optional default context hold by Tera.
Thus we can then create Context with some default values already filled and avoid redundant code.


Keats:
This has already been added in https://github.com/Keats/tera2 (https://github.com/Keats/tera2/blob/master/tera/src/tera.rs#L623-L647). I'm not going to add it to Tera 1

PR #964, opened 2025-03-18
Docs: Add note about child templates ignoring code outside of a block

Raymi306:
My original conversation: https://zola.discourse.group/t/question-importance-of-tera-variable-assignment-location-not-loop-related/2525

I am certainly open to suggestions for wording as well.


Keats:
Seems fine, thanks!

PR #967, opened 2025-04-19
Allow sort filter to be case insensitive

### LINKED ISSUE #587: Case-insensitive sort

### LINKED ISSUE #831: Add a case_sensitive attribute to the sort filter
When sorting strings sometimes one doesn't care whether a letter is uppercase or lowercase just that the 'a's are near the 'A's.

The current behavior of `sort`  is that it sorts (for Latin at least) Uppercase before lowercase:
* A_1
* B_3
* a_2

When someone would implement this the resulting `sort(case_sensitive=false)` would sort as if the attribute to sort by was ran through a `lowercase` filter first:
* A_1
* a_2
* B_3

If that is too complex for the current version of tera because of implementation details I'm not aware of please consider it for the v2 wishlist.

doamatto:
I saw it was proper ages ago that someone tried bringing case sensitivity to the `sort` filter and figured I'd go at it. If I'm not mistaken, Tera will use case sensitivity by default (that is, prior to and after this PR) so I only added a test for case insensitivity.

Closes #587
Closes #831 
See also : #693


cyckl:
Thank you so much for writing this patch—I've been looking for this functionality for ages but haven't had the time to implement something myself!

doamatto:
Wanted to throw in a wee screenshot too of it behaving as expected, as well :

image


<details open>
<summary>Before</summary>

image


</details>

Keats:
I think it's fine to merge as is and change the MSRV after merge.
By the way I haven't implemented the sort filter yet in tera2 (https://github.com/Keats/tera2) if someone is interesting in porting it.

doamatto:
> By the way I haven't implemented the sort filter yet in tera2 (https://github.com/Keats/tera2) if someone is interesting in porting it.

I might have some time to work on it this month — are there any quirks or other things I might need to know about with the new version ?

Keats:
It should be easier to implement, they look more like regular functions. For example: https://github.com/Keats/tera2/blob/master/tera/src/filters.rs#L322-L341

azemlya:
> It should be easier to implement, they look more like regular functions. For example: https://github.com/Keats/tera2/blob/master/tera/src/filters.rs#L322-L341

Since the round topic has been touched upon, I will make my suggestions regarding this filter. It would be logical to add the trunc method.
```
...
    Some("trunc") => Ok(((multiplier * val).trunc() / multiplier).into()),
...
```

Keats:
Sure that can be added

Keats:
Go for it

Issue #974, opened 2025-09-09
How to enable `strict` mode?


josecelano:
The documentation mentions that there is a "strict" mode:

<img width="567" height="414" alt="Image" src="https://github.com/user-attachments/assets/abee91a2-c4d2-45fd-ab5f-650794083569" />

However, I do not know if:

- It's the default mode
- If not, how can I enable it?

I would like to show a warning with "tracing" when the context has more variables than what the template requires.

Kek5chen:
Stumbling over this too right now. For my current case I do not care for the existance, and I do not want to force the user to provide a `default` filter, as that is bothersome, and even error prone as it fails to render the template. An empty string replacement would do the trick.

Kek5chen:
I can't find anything about a toggle like this though. I'll see if I can make a PR for us.

Keats:
I wouldn't accept something a PR like that though. Erroring on undefined things is intended and it not planned to add a toggle.

As for the screenshot, I'm not sure which documention it's coming from but there has never been a strict mode you can set.

Kek5chen:
@josecelano gladly also give it a try and tell me if there's any issues ^^

Keats:
It's just bad wording. It means it is strict as opposed to some template engines that will not care about undefined

KrullBorg:
> It's just bad wording. It means it is strict as opposed to some template engines that will not care about undefined

but here [tera doc](https://keats.github.io/tera/docs/#if) you talk about "Undefined variables are considered falsy": i tried and if i don't define a variable and i test it (boolean) i don't get an error (as expected)

instead i get an error if i try just use it normally (not as boolean but not defined in context)

the problem is that an undefined and a defined but empty variable return false in the same way

maybe i didn't understand something

thanks in advance

PR #976, opened 2025-10-07
Opt-Out strict mode

Kek5chen:
As mentioned in #974, opting out of Teras strictness would help out some people with their work.

This PR is made to fix a personal issue of mine and others, in turn also making Tera more flexible.

I know the spirit of Tera is to be safe and I've tried to uphold everything in this PR to be non-breaking, so that only those who wish to explicitly opt out of the strict behavior can do so.

Tests and documentation are provided and ran through successfully.

I really hope this can become a part of Tera, as currently no other library provides this flexibility, except the usual (Golang) equivalents. Rust desperately needs this...

I can assert full ownership of the patch. No AI was used.

### DIFF
--- a/src/renderer/processor.rs
+++ b/src/renderer/processor.rs
@@ -400,11 +400,14 @@ impl<'a> Processor<'a> {
                         if expr.has_default_filter() {
                             self.get_default_value(expr)?
                         } else {
-                            if !expr.negated {
-                                return Err(e);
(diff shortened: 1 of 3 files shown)


Keats:
That's not the right approach. Something like https://github.com/Keats/tera2/issues/72#issuecomment-3374088513 is much more likely to happen

PR #979, opened 2025-10-21
refactor: 📦 Replace unmaintained unic-segment dependency

### LINKED ISSUE #978: Rustsec advisory flags tera via unic
[A Rustsec advisory ](https://rustsec.org/advisories/RUSTSEC-2025-0095.html) was recently published that flags `unic` as unmaintained.

That however flags tera (and upstream crates) on CI that use `cargo deny` and similar. 

Image

Would it be possible to release tera 1.21 with updated dependencies? The advisory suggests one of

- [icu](https://crates.io/crates/icu)
- [idna](https://crates.io/crates/idna)
- [unicode-bidi](https://crates.io/crates/unicode-bidi)

berkus:
Fixes #978

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
           os: macOS-latest
           rust: stable
         - build: win-msvc
-          os: windows-2019
+          os: windows-2022
           rust: stable
     steps:
     - uses: actions/checkout@v5


berkus:
@Keats alright, bumped it, passes locally.

Keats:
looks like the embark actions need some updates. I think they don't use Rust anymore, is there another one?

PR #980, opened 2025-10-30
docs: describe allowed characters in names

### LINKED ISSUE #913: Macro names can have underscores but not hyphens
[The Tera macro documentation][1] is silent on this matter (to my knowledge), but I've verified that macro names can have underscores but not hyphens. I think it would be nice to state this in the documentation.

[1]: https://keats.github.io/tera/docs/#macros

ojob:
This applies to declared variables, including in for loops, and macros.

Closes: #913


docs/content/docs/_index.md
```diff
@@ -352,6 +352,8 @@ in for loop are valid until the end of the current iteration only.
 {% set my_var = [1, true, some_var | round] %}
 ```
 
+Note that variables names are limited to ASCII letters, figures, and underscore, i.e. `[a-zA-Z0-9_]`.
```
Keats: do you mean numbers rather than figures? They also can't start with a number

PR #993, opened 2026-06-29
Added get_component_names and contains_component_definition.

tobiastom:
I needed to get the names of all components that are registered. This pull request adds similar methods to the ones that exist for getting templates without exposing the components. 

Hope this looks good to you.

### DIFF
--- a/tera/src/tera.rs
+++ b/tera/src/tera.rs
@@ -428,6 +428,30 @@ impl Tera {
             .map(|(def, _)| ComponentInfo::from(def))
     }
 
+    /// Lookups a component by name, returning whether it's found or not
+    pub fn contains_component_definition(&self, component_name: &str) -> bool {
+        self.get_component_definition(component_name).is_some()
+    }
+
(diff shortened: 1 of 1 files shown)


tera/src/tera.rs
```diff
@@ -428,6 +428,30 @@ impl Tera {
             .map(|(def, _)| ComponentInfo::from(def))
     }
 
+    /// Lookups a component by name, returning whether it's found or not
+    pub fn contains_component_definition(&self, component_name: &str) -> bool {
```
Keats: `contains_component_definition` -> `contains_component`.

Can you add an example as well? It can be almost copied from the one below

Keats (changes requested):

tobiastom:
@Keats Sorry, I didn't find the time to fix it myself. Thank you very much! 

btw, you've written `fakse` instead of `false` in the first line. 🫣

Keats:
> btw, you've written fakse instead of false in the first line. 🫣

:(

PR #1000, opened 2026-07-07
Add .iter() method to Kwargs

### LINKED ISSUE #999: Can't Iterate Over `tera::Kwargs`
This might also belong under #992 , but figured I'd put it here so it could be tracked easier.

In Tera V1, because the arguments to a registered function was just a `HashMap<String, tera::Value>`, we could have code [like this](https://github.com/p0lyw0lf/driver/blob/215fec9b64ab4c7dfabb1d6ccbebec1f15ff2cf6/packages/driver-query-ssg/src/tera/mod.rs#L354-L372) that iterated over all the kwargs to build another object that gets passed elsewhere.

In Tera V2, we cannot access the underlying `Map` object, nor does the `Kwargs` object expose a method to iterate over its internals, so we're stuck.

For the sake of forwards-compatibility, something like the following should work:

```rust
impl Kwargs {
    pub fn iter(&self) -> impl Iterator<Item = (&Key<'static>, &Value)> {
        self.values.iter()
    }
}
```

If you think this is a good idea, I can open a PR with this.

p0lyw0lf:
Resolves #999.

### DIFF
--- a/tera/src/args.rs
+++ b/tera/src/args.rs
@@ -332,6 +332,12 @@ impl Kwargs {
             Err(Error::missing_arg(key))
         }
     }
+
+    /// Iterates over all the provided arguments. Order is not guaranteed unless the
+    /// "preserve_order" feature is set.
+    pub fn iter(&self) -> impl Iterator<Item = (&Key<'static>, &Value)> {
+        self.values.iter()
+    }
 }
 
 impl<const N: usize> From<[(&'static str, Value); N]> for Kwargs {


antoine-de:
Do you know when this will be released? I need this for the fluent-template (https://github.com/XAMPPRocky/fluent-templates/pull/108) migration

Keats:
Probably soon, I was waiting to see if there are more things to fix so I can batch them in a single release

PR #1010, opened 2026-08-09
Support negative component default values

### LINKED ISSUE #1009: Can't use negative numbers as component defaults.
When specifying components one can set default values which works fine like this:

```
{% component test(foo=0) %}
	{{foo}}
{% endcomponent test %}
```

But trying to put a negative integer (or float) in its place fails:

```
{% component test(foo=-1) %}
	{{foo}}
{% endcomponent test %}
```

```
Failed to load /[…]/templates/tera/helpers.html: error: Found `-` but component default arguments can only be one of: string, bool, integer, float, array or map
      --> helpers.html:20:23
       |
    20 | {% component test(foo=-1) %}
       | 
```

Similar error when wrapping it in `{-1}`.

oiahoon:
### Changes

- accept negative integer and float literals as component definition defaults
- preserve each default's value and inferred argument type
- add rendering and introspection regression coverage

Fixes #1009.

### Verification

- `cargo fmt --all --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test` (115 passed)
- `cargo test --all-features` (171 passed)
- focused rendering and introspection regressions
- `git diff --check`

### DIFF
--- a/tera/src/parsing/parser.rs
+++ b/tera/src/parsing/parser.rs
@@ -14,7 +14,7 @@ use crate::parsing::ast::{
 use crate::parsing::ast::{BinaryOperator, Node, UnaryOperator};
 use crate::parsing::lexer::{Token, tokenize};
 use crate::utils::{Span, Spanned};
-use crate::value::{Key, Value};
+use crate::value::{Key, Value, ValueInner};
 use crate::{HashMap, HashSet};
 
 /// Maximum recursion depth for the parser, shared between expression and statement parsing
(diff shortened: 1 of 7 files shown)


Keats:
Thanks! One thing we need to ensure is that `[1]` and `[-1]` and `{"foo": -1}` work as well for component default params

oiahoon:
Addressed in 49ddff5. Added coverage for positive and negative arrays plus maps containing negative values.

Keats (changes requested):
I think it would be better to do constant folding for unaries at parse time so everything down the line does not need to be changed/special case it

tera/src/parsing/parser.rs
```diff
@@ -731,7 +731,27 @@ impl<'a> Parser<'a> {
                 let (_, r_bp) = unary_binding_power(op);
                 let expr = self.inner_parse_expression(r_bp)?;
                 span.expand(&self.current_span);
-                Expression::UnaryOperation(Spanned::new(UnaryOperation { op, expr }, span.clone()))
+                match (op, &expr) {
+                    (UnaryOperator::Minus, Expression::Const(value)) => match value.node().inner {
+                        ValueInner::I64(value) => Expression::Const(Spanned::new(
+                            Value::from(value.checked_neg().ok_or_else(|| {
+                                Error::syntax_error("Integer overflow".to_string(), &span)
+                            })?),
```
Keats, line 739: Can that error ever happen?

tera/src/parsing/parser.rs
```diff
@@ -1318,6 +1338,19 @@ impl<'a> Parser<'a> {
...
                     Some(Ok((Token::Integer(b), _))) => (Value::from(*b), true),
                     Some(Ok((Token::Float(b), _))) => (Value::from(*b), true),
+                    Some(Ok((Token::Minus, _))) => {
+                        self.next_or_error()?;
+                        match &self.next {
+                            Some(Ok((Token::Integer(b), _))) => (Value::from(-*b), true),
+                            Some(Ok((Token::Float(b), _))) => (Value::from(-*b), true),
+                            _ => {
+                                return Err(Error::syntax_error(
+                                    "Found `-` but component default arguments can only be one of: string, bool, integer, float, array or map".to_string(),
```
Keats, line 1348: Can you add a snapshot error for that error? eg a component getting `-true` as default argument. Or I can do it after merging, up to you

Keats:
Thanks!

PR #1012, opened 2026-08-12
Add configurable indent character

### LINKED ISSUE #1007: Feature request: indent by arbitrary character
[`indent`](https://keats.github.io/tera/#indent) built-in filter always uses a whitespace (U+20), but sometimes other characters such as tab (U+9) is used as an indent character. Current `indent` is not useful for such context (especially usual for XML and HTML documents).

It would be great if non-U+20 characters can be used as a indent character by passing as an argument, for example `indent(width=1, char='\t')`. Default `width` of 4 would be too much for tab character, but I feel it's acceptable to always explicitly specify `width=1` when using tab character.

ColumbusLabs:
Closes #1007

Adds an optional `char` argument to the built-in `indent` filter while preserving the existing four-space default. The value must contain exactly one Unicode character, so the existing width cap continues to bound output expansion.

The rendering snapshot covers tab indentation, unit tests cover a non-ASCII character plus empty/multi-character rejection, and the filter reference documents the new option.

Validation:
- `cargo fmt --all --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test`
- `cargo test --all-features`
- `zola build`

### DIFF
--- a/docs/content/_index.md
+++ b/docs/content/_index.md
@@ -902,7 +902,7 @@ need to call the `safe` filter after `newlines_to_br`.
 ##### indent
 Indents a string by injecting a prefix at the start of each line.
 The `width` argument (default `4`) specifies how many indent characters to insert per line.
-The `char` argument (default `" "`) specifies the indent character, such as `"\t"` for a tab.
+The `indentation` argument (default `" "`) specifies the indent character, such as `"\t"` for a tab.
 If the `first` argument (default `false`) is set `true`, the prefix is inserted for the first line.
 If the `blank` argument (default `false`) is set `true`, the prefix is inserted for blank/whitespace lines.
 
--- a/docs/content/_index.md
+++ b/docs/content/_index.md
@@ -901,9 +901,10 @@ need to call the `safe` filter after `newlines_to_br`.
 
 ##### indent
 Indents a string by injecting a prefix at the start of each line.
-The `width` argument (default `4`) specifies how many spaces to insert per line.
-If the `first` argument (default `false`) is set `true`, spaces are inserted for the first line.
-If the `blank` argument (default `false`) is set `true`, spaces are inserted for blank/whitespace lines.
+The `width` argument (default `4`) specifies how many indent characters to insert per line.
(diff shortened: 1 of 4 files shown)


Keats (approved):

Keats:
Looks great thanks! I'll merge when I am ready to release the next version as it updates the doc site and don't want to mislead for now. New version should be releases soonish if I find the time to fix some issues other people mentioned

lo48576:
It was totally my bad to name it `char` in #1007, but it seems like any strings can be used as an indent, so maybe it should have a different name, like `str`, `indent`, or something?

The argument accepting a string rather than a single character looks reasonable to me, since it makes it possible to use "fancy" indents, such as `&nbsp;&nbsp;` and `<span class='indent'>    </span>`.

Keats:
Yeah sounds good for renaming the parameter. I'd pick `indentation` myself

lo48576:
> but it seems like any strings can be used as an indent,

https://github.com/ColumbusLabs/tera/blob/1566e49ab8c3ec79c31ee3b7520adf6cb907f5cc/tera/src/filters.rs#L254-L258

Ah sorry I missed that, it checks that `char` is actually a single char. Then this part should be omitted if we allow indentation by string.

ColumbusLabs:
To make sure I apply the intended behavior: should `indentation` accept any non-empty string (for example `&nbsp;`), or should the renamed argument continue to require one Unicode character? I can update the implementation and tests either way.

Keats:
It depends whether we actually want to support more than one char for indentation. In practice, I don't really see a need for it but we could rename char -> indentation so if we decide to handle it later we can reuse the param

ColumbusLabs:
Updated the parameter to `indentation` while keeping the existing single-Unicode-character validation. Docs, the rendering fixture, and unit coverage now use the new name. `cargo fmt --check` and `cargo test -p tera` passed (72 unit tests, 15 integration tests, and 30 doctests).

Keats:
Thanks!

PR #1013, opened 2026-08-12
Add chained filter sections

### LINKED ISSUE #757: Chained filter sections
It would be nice if a single filter section could chain multiple filters instead of having to nest the sections inside each other.

e.g. instead of
```tera
{% filter highlight(lang="toml") %}{% filter dedent -%}
    [package.metadata.docs.rs]
    rustc-args = ["--cfg", "docsrs"]
{%- endfilter %}{% endfilter %}
```

it could be

```tera
{% filter dedent | highlight(lang="toml") -%}
    [package.metadata.docs.rs]
    rustc-args = ["--cfg", "docsrs"]
{%- endfilter %}
```

ebkalderon:
Adds support for chaining multiple filters in a single `{% filter %} ... {% endfilter %}` section, with each filter delimited by the `|` character. For instance, the following:

```jinja2
{% filter safe %}{% filter replace(from="Robert", to="Bob") %}{% filter indent %}
Hello, I'm Robert!
{% endfilter %}{% endfilter %}{% endfilter %}
```

may now be rewritten as:

```jinja2
{% filter indent | replace(from="Robert", to="Bob") | safe %}
Hello, I'm Robert!
{% endfilter %}
```

Closes #757.

Looking forward to using this feature while tackling https://github.com/ebkalderon/terminus/issues/27.

### DIFF
--- a/tera/src/parsing/ast.rs
+++ b/tera/src/parsing/ast.rs
@@ -2,7 +2,6 @@ use std::collections::{BTreeMap, HashSet};
 use std::fmt;
 use std::str::FromStr;
 
-use crate::HashMap;
 use crate::errors::Error;
 use crate::utils::{Span, Spanned};
 use crate::value::{Key, Value, ValueInner, format_map};
@@ -255,11 +254,11 @@ impl fmt::Display for Expression {
     }
 }
 
-#[derive(Clone, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct Filter {
     pub expr: Expression,
     pub name: String,
-    pub kwargs: HashMap<String, Expression>,
+    pub kwargs: BTreeMap<String, Expression>,
 }
 
 impl fmt::Display for Filter {
@@ -277,21 +276,6 @@ impl fmt::Display for Filter {
             }
         }
         write!(f, "}})",)
-    }
-}
-
-impl fmt::Debug for Filter {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let Self { expr, name, kwargs } = &self;
-
-        // Ensures `Debug` output is stable in `insta` snapshot tests
-        let sorted_kwargs: BTreeMap<_, _> = kwargs.iter().collect();
-
-        f.debug_struct(stringify!(Filter))
-            .field("expr", expr)
-            .field("name", name)
-            .field("kwargs", &sorted_kwargs)
-            .finish()
     }
 }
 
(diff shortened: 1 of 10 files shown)


tera/src/parsing/ast.rs
```diff
@@ -280,6 +280,21 @@ impl fmt::Display for Filter {
     }
 }
 
+impl fmt::Debug for Filter {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let Self { expr, name, kwargs } = &self;
+
+        // Ensures `Debug` output is stable in `insta` snapshot tests
+        let sorted_kwargs: BTreeMap<_, _> = kwargs.iter().collect();
```
Keats: We should probably store them in a btreemap in the first place eh (no need to change anything)

tera/src/snapshot_tests/rendering_inputs/success/filter_section.txt
```diff
@@ -0,0 +1 @@
+{% filter replace(from="Robert", to="Bob") | safe -%} Hello I'm Robert! {%- endfilter %}
```
Keats: maybe let's use something like `upper` instead of `safe` to make sure we see it's doing something.

Keats (changes requested):

tera/src/parsing/ast.rs
ebkalderon: No worries! I'm more than happy to switch the existing `HashMap` out for a `BTreeMap`, if desired. Would you consider this change to be a breaking API change, though?

tera/src/snapshot_tests/rendering_inputs/success/filter_section.txt
ebkalderon: In this case, the `safe` filter unescapes the `'` character. Still, I can certainly swap it out for `upper` filter instead so the effect is more visually noticeable. 👌

tera/src/parsing/ast.rs
Keats: We don't expose the AST. But I'll do it myself later if needed, no need to do it now

ebkalderon: It's all good, haha. I've made the change in https://github.com/Keats/tera/pull/1013/commits/e93f69ad1b99f8a7661b838e39761f0332e1fd2a. Does this look good to you?

tera/src/snapshot_tests/rendering_inputs/success/filter_section.txt
ebkalderon: Okay, I've added `upper` to this snapshot test to make the rendered end-result more visually striking! However, I still kept the `safe` filter at the end, since this escapes the `'` character and turns the final output into `HELLO I'M BOB!` instead of `HELLO I&#39;M BOB!`. Does https://github.com/Keats/tera/pull/1013/commits/c79f9cf2b35d0645dab3f9faa89aad3a371bc1bc look acceptable to you?

Keats (approved):

PR #1019, opened 2026-08-18
Expose escape fn on State

### LINKED ISSUE #1011: Feature request: Access escape function from function `State`
While implementing my localization system I ran into the problem, that sometimes functions output safe text, but in order to do so need to escape their inputs. For this usecase it would be great if the function could access the escaping function of the template directly instead of having to make guesses.

Keats:
Closes #1011

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,16 @@
 # Changelog
 
+## 2.3.0 (unreleased)
+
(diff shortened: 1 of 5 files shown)


slatian:
That looks like it will solve my problem, Big Thank You!

I have not found the time to test it this weekend though.

Keats:
https://github.com/Keats/tera/pull/1024 is using that so if it's enough for Tera itself it's probably enough for 3rd parties

PR #1024, opened 2026-08-22
Escaping

Keats:


Keats:
@ebkalderon turns out it was more involved to fix the filter section escaping...
I'll double check it next week to make sure I haven't made any mistakes

PR #1032, opened 2026-09-01
Explicitly specify the serde/derive dev-dependency

musicinmybrain:
In downstream packaging for Fedora, we routinely patch out dev-dependencies on `criterion` because they are unnecessary (we run tests, but not benchmarks) and bring in a large tree of otherwise-unnecessary dependencies.

It turns out that dropping the `criterion` dev-dependency by removing this line from `Cargo.toml`

https://github.com/Keats/tera/blob/f345cf2370456a868a5074ad2372ff4db8ecbcc0/tera/Cargo.toml#L45

breaks compiling the tests and examples, with many errors like:

```
error: cannot find derive macro `Serialize` in this scope
 --> examples/teams.rs:5:10
  |
5 | #[derive(Serialize)]
  |          ^^^^^^^^^
  |
note: `Serialize` is imported here, but it is only a trait, without a derive macro
 --> examples/teams.rs:1:5
  |
1 | use serde::Serialize;
  |     ^^^^^^^^^^^^^^^^
help: consider importing this derive macro
  |
1 + use serde_derive::Serialize;
  |
```

This is because tests and examples rely on the `derive` feature of the `serde` crate (not just having `serde_derive` in the dependency tree, although the explicit dependency on it is correct since `serde_derive::Serialize` is used directly), but there’s no (dev-)dependency on `serde` with the `derive` feature. Instead, `criterion` happens to depend on `serde/derive`, so it gets enabled due to [feature unification](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification).

To be more explicit about this requirement, to guard against possible changes in the features used by `criterion` or perhaps a future switch to a different benchmarking library, and to make downstream packaging easier, this PR adds an explicit dev-dependency on `serde/derive`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1210,7 +1210,6 @@ dependencies = [
  "itoa",
  "pulldown-cmark-escape",
  "serde",
- "serde_derive",
  "tempfile",
  "unicode-segmentation",
  "walkdir",
--- a/tera/Cargo.toml
+++ b/tera/Cargo.toml
@@ -45,7 +45,6 @@ glob_fs = ["dep:walkdir", "dep:globset"]
 criterion = "0.8"
 insta = { version = "1", features = ["glob"] }
 serde = { version = "1", features = ["derive"] }
-serde_derive = "1.0.156"
 tempfile = "3"
 
 [package.metadata.docs.rs]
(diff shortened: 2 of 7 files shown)


musicinmybrain:
@decathorpe pointed out that with the `serde/derive` dev-dependency properly specified, it’s no longer necessary to use and depend on `serde_derive` directly, and that this would eliminate the need to update `serde` and `serde_derive` in lockstep. I pushed a follow-up commit implementing that suggestion.

Keats:
Thanks!

PR #1037, opened 2026-09-07
Introducing positional arguments in functions, filters and tests

oschijns:
I would like to use Tera to replace Jinja in a project, but my Jinja templates contain a lot of functions using positional arguments instead of named ones.
The goal of this PR would be to support both.

I have 7 non passing tests, and I guess I will have to write new ones for the new features.

What do you think?

### DIFF
--- a/tera/src/args.rs
+++ b/tera/src/args.rs
@@ -334,6 +334,31 @@ impl Kwargs {
         }
     }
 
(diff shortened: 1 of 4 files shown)


Keats:
Nope it's not planned. You have https://github.com/mitsuhiko/minijinja if you want something exactly like jinja2

PR #1038, opened 2026-09-08
Guard Compiler::compile_expr's recursion depth

### LINKED ISSUE #1035: Unbounded recursion in Compiler::compile_expr aborts tera on deep expression chains (stack overflow)
## Summary

`Compiler::compile_expr` recurses into the AST with no depth limit, while the parser's own recursion guard (`MAX_RECURSION_DEPTH = 40`) doesn't cover this path. The parser builds left-associative chains (`1 + 1 + …`, `a | f | f | …`, `a.b.c…`) *iteratively* in its Pratt loop, so `recursion_depth` never increments for them and the guard never fires. The compiler then walks that same left-leaning AST *recursively*:

```rust
// parsing/compiler.rs, compile_expr
Expression::BinaryOperation(op) => {
    self.compile_expr(op.left)?;   // line 414
    self.compile_expr(op.right)?;  // line 415
    ...
}
Expression::Filter(filter) => { self.compile_expr(filter.expr)?; ... }   // line 224
Expression::GetAttr(attr)  => { self.compile_expr(attr.expr)?; ... }     // line 177
```

There's no counter in `Compiler` at all. For a chain of `n` terms this recurses `n` deep down the left spine and overflows the Rust stack — a hard `fatal runtime error: stack overflow` (SIGABRT), not a catchable `panic!`, so `catch_unwind` doesn't help. This is a syntactically valid template (`{{ 1 + 1 + … + 1 }}`), not malformed input — a 5,000-term chain of the same shape renders fine.

## Repro

Self-contained PoC against `tera = { git = "...", rev = "1381ea5" }` (`v2.1.1`, and reproduces identically on `main` HEAD `98551d2`):

```
$ cargo run --release -- crash 100000      # {{ 1 + 1 + ... + 1 }}, 400,007 bytes
thread 'main' has overflowed its stack
fatal runtime error: stack overflow, aborting
exit=134

$ cargo run --release -- filter 100000     # {{ 1 | upper | upper | ... }}, 800,007 bytes
exit=134

$ cargo run --release -- attr 100000       # {{ a.b.b.b... }}, 200,007 bytes
exit=134

$ cargo run --release -- crash 5000        # below threshold
OK (rendered)                              # exit=0
```

Threshold on an 8 MiB main-thread stack: `n=6000` succeeds, `n=6500` (~26 KB of input) aborts.

Controls:

| Case | Result |
|---|---|
| `crash 5000` — 20 KB valid chain | renders OK, exit 0 |
| `benign 100000` — ~600 KB plain-content template | renders OK, exit 0 |
| `paren 100000` — `{{ ((((…1…)))) }}`, right-*nested* not left-*chained* | rejected gracefully: `The expression is too complex`, exit 0 |
| `crash 100000` / `filter 100000` / `attr 100000` — left-leaning chains | abort, exit 134 |

The `paren` control shows the parser's guard does work for right-nested forms; the left-leaning chains just never reach it, and nothing on the compiler side stands in for it.

## Reachability

`Tera::render_str` (documented for "a template coming from user input"), `Tera::one_off`, and the file/glob loaders all funnel into `Template::new` → `Parser::parse` → `Compiler::compile` → `compile_expr`. The crash happens during compilation, before any reference validation or rendering — no undefined variable, custom filter, or function is required.

Happy to open a PR adding a depth budget to `compile_expr` (mirroring the parser's `recursion_depth`) if useful.

zikk090:
Fixes #1035. The parser already bounds its own recursion via MAX_RECURSION_DEPTH, but that guard only fires for right-nested forms built through inner_parse_expression (parentheses, unary operators, **). Left-associative chains -- `1 + 1 + ...`, `a | f | f | ...`, `a.b.c...` -- are built iteratively by the Pratt loop and never increment recursion_depth, so they sail past the parser untouched. Compiler::compile_expr then walks that same left-leaning AST recursively with no guard of its own, so a template with a few thousand chained terms aborts the process with a stack overflow instead of returning an error.

This adds the same MAX_RECURSION_DEPTH guard to compile_expr (now Compiler::compile_expr, wrapping the renamed compile_expr_inner), returning the same "The expression is too complex" error the parser already uses for right-nested forms. Making compile_expr fallible means propagating TeraResult through everything that calls it, directly or transitively: compile_kwargs, compile_map_entries, compile_block, compile_node, and Compiler::compile, plus their two call sites in Template::new.

Second bug found and fixed while verifying this: once the depth guard returns early, the still-mostly-intact remainder of the expression tree it was handed needs to be dropped -- and Expression's default, compiler-derived Drop is itself recursive over that same boxed tree, so a sufficiently large chain (tested: ~55,000+ terms) traded a stack overflow during *compilation* for one during *drop*, on the very error path meant to prevent the crash. Expression::drop_iteratively walks the tree with an explicit heap-allocated stack instead of the call stack, and compile_expr's guard now calls it explicitly on the rejected expression instead of letting it drop implicitly. Verified directly: templates up to 5,000,000 chained terms (about 20MB of source) now return a clean error with no crash, versus overflowing before this second fix at around 55-60k terms.

Tested: `cargo test` and `cargo test --all-features` both pass (139 tests across all_features), `cargo fmt --all --check` and `cargo clippy --all-targets --all-features -- -D warnings` both clean. Verified against the exact reproduction from #1035 plus a range of chain lengths from 100 to 5,000,000 terms to find and confirm both the original threshold and the drop-related one.

### DIFF
--- a/tera/src/parsing/ast.rs
+++ b/tera/src/parsing/ast.rs
@@ -166,6 +166,122 @@ impl Expression {
             Expression::ListComprehension(s) => s.span_mut().expand(span),
         }
     }
+
+    /// Drops `self` iteratively instead of relying on the default, recursive `Drop` glue.
+    ///
+    /// `Expression` is a boxed recursive tree (each nested expression sits behind a `Box` in
+    /// its `Spanned<T>` wrapper), so the compiler-generated `Drop` implementation walks it
+    /// recursively -- one stack frame per level. For a sufficiently deep left-leaning chain
+    /// (`1 + 1 + ... + 1`), simply *dropping* the tree can overflow the stack on its own, even
+    /// with nothing else recursing over it.
+    ///
+    /// This matters specifically because `Compiler::compile_expr`'s recursion guard (see
+    /// `parsing/compiler.rs`) can return an error *before* fully consuming a deeply nested
+    /// `Expression` it was handed -- at that point the still-mostly-intact remainder of the
+    /// tree would otherwise be dropped in one shot by the caller. Call this method instead of
+    /// letting that happen implicitly.
+    ///
+    /// This does not need to be (and is not) a general `Drop` impl: every other place an
+    /// `Expression` is consumed already walks it one level at a time via `compile_expr`'s own
+    /// (now depth-guarded) recursion, so the tree is always shallow by the time it would
+    /// naturally go out of scope.
+    pub(crate) fn drop_iteratively(self) {
+        let mut stack = vec![self];
+        while let Some(expr) = stack.pop() {
+            match expr {
+                Expression::BinaryOperation(e) => {
+                    let (op, _) = e.into_parts();
+                    stack.push(op.left);
+                    stack.push(op.right);
+                }
+                Expression::UnaryOperation(e) => {
+                    let (op, _) = e.into_parts();
+                    stack.push(op.expr);
+                }
+                Expression::GetAttr(e) => {
+                    let (attr, _) = e.into_parts();
+                    stack.push(attr.expr);
+                }
+                Expression::GetItem(e) => {
+                    let (item, _) = e.into_parts();
+                    stack.push(item.expr);
+                    stack.push(item.sub_expr);
+                }
+                Expression::Slice(e) => {
+                    let (slice, _) = e.into_parts();
+                    stack.push(slice.expr);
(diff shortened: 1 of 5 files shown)


Keats:
Thanks but I don't think this is the correct approach. I've pushed https://github.com/Keats/tera/pull/1040 instead

